PR #6, opened 2019-04-05
Allow creating seeds with the install command

kennyadsl:
This PR allows to call:

```
STRIPE_SEEDS_SECRET_KEY=xxx \
STRIPE_SEEDS_PUBLISHABLE_KEY=yyy \
bundle exec rails g solidus_stripe:install --migrate=false --auto_run_seeds=true
```

to run seeds along with the engine installation. This could be useful in the to allow using stripe in the core sandbox if we decide to go on that path for https://github.com/solidusio/solidus/issues/3162.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+secret_key = ENV['STRIPE_SEEDS_SECRET_KEY']
+publishable_key = ENV['STRIPE_SEEDS_PUBLISHABLE_KEY']
+
+puts "Loading seed: solidus_stripe/stripe_payment_method"
+
+if secret_key.blank? || publishable_key.blank?
+  puts "Failure: You have to set both STRIPE_SEEDS_SECRET_KEY and STRIPE_SEEDS_PUBLISHABLE_KEY environment variables."
+else
+  stripe_payment_method = Spree::PaymentMethod::StripeCreditCard.new do |payment_method|
+    payment_method.name = 'Credit Card'
+    payment_method.preferred_test_mode = true
+    payment_method.preferred_secret_key = secret_key
+    payment_method.preferred_publishable_key = publishable_key
+  end
+
+  if stripe_payment_method.save
+    puts "Stripe Payment Method correctly created."
+  else
+    puts "There was some problems with creating Stripe Payment Method:"
+    stripe_payment_method.errors.full_messages.each do |error|
+      puts error
+    end
+  end
+end
--- a/lib/generators/solidus_stripe/install/install_generator.rb
+++ b/lib/generators/solidus_stripe/install/install_generator.rb
@@ -1,18 +1,29 @@
+# frozen_string_literal: true
+
 module SolidusStripe
   module Generators
     class InstallGenerator < Rails::Generators::Base
+      class_option :migrate, type: :boolean, default: true
       class_option :auto_run_migrations, type: :boolean, default: false
+      class_option :auto_run_seeds, type: :boolean, default: false
 
       def add_migrations
         run 'bundle exec rake railties:install:migrations FROM=solidus_stripe'
       end
 
       def run_migrations
-        if running_migrations?
+        if options.migrate? && running_migrations?
           run 'bundle exec rake db:migrate'
         else
           puts "Skiping rake db:migrate, don't forget to run it!"
         end
+      end
+
+      def populate_seed_data
+        return unless options.auto_run_seeds?
+
+        say_status :loading, 'stripe seed data'
+        rake('db:seed:solidus_stripe')
       end
 
       private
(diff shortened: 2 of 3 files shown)


lib/generators/solidus_stripe/install/install_generator.rb
```diff
@@ -1,20 +1,37 @@
...
           puts "Skiping rake db:migrate, don't forget to run it!"
         end
       end
 
+      def populate_seed_data
+        return unless options.auto_run_seeds?
+
+        say_status :loading, "stripe seed data"
+        rake_options = []
+        rake_options << "STRIPE_SEEDS_SECRET_KEY=#{options[:stripe_seeds_secret_key]}" if options[:stripe_seeds_secret_key]
```
tvdeyen: When we already use env vars we don’t need to have to pass them to rake like this.

lib/generators/solidus_stripe/install/install_generator.rb
```diff
@@ -1,20 +1,37 @@
+# frozen_string_literal: true
+
 module SolidusStripe
   module Generators
     class InstallGenerator < Rails::Generators::Base
+      class_option :migrate, type: :boolean, default: true
       class_option :auto_run_migrations, type: :boolean, default: false
+      class_option :auto_run_seeds, type: :boolean, default: false
+      class_option :stripe_seeds_secret_key, type: :string
```
tvdeyen: We don’t need to have extra options when we already use env vars in the seeds file.

db/seeds.rb
```diff
@@ -0,0 +1,27 @@
...
+
+puts "Loading seed: solidus_stripe/stripe_payment_method"
+
+if secret_key.blank? || publishable_key.blank?
+  puts "Failure: You have to set both STRIPE_SEEDS_SECRET_KEY and STRIPE_SEEDS_PUBLISHABLE_KEY environment variables."
+else
+  stripe_payment_method = Spree::PaymentMethod::StripeCreditCard.new do |gateway|
+    gateway.name = 'Credit Card'
+    gateway.preferred_server = "test"
+    gateway.preferred_test_mode = true
```
tvdeyen: Should we make this 

```rb
gateway.preferred_test_mode = !Rails.env.production?
```

db/seeds.rb
```diff
@@ -0,0 +1,27 @@
...
+secret_key = ENV['STRIPE_SEEDS_SECRET_KEY']
+publishable_key = ENV['STRIPE_SEEDS_PUBLISHABLE_KEY']
+
+puts "Loading seed: solidus_stripe/stripe_payment_method"
+
+if secret_key.blank? || publishable_key.blank?
+  puts "Failure: You have to set both STRIPE_SEEDS_SECRET_KEY and STRIPE_SEEDS_PUBLISHABLE_KEY environment variables."
+else
+  stripe_payment_method = Spree::PaymentMethod::StripeCreditCard.new do |gateway|
+    gateway.name = 'Credit Card'
```
tvdeyen: This is not a `gateway`,  but a `payment_method` 🤓

tvdeyen (commented):
This is nice. Maybe we could make this ready for seeding production as well. And I think we do not need the extra options, as we already use env vars.

db/seeds.rb
kennyadsl: I think in production it is better to use `static_model_preferences` so making this available in production as well could encourage a bad practice. 

Unfortunately, I can't figure out a way to use `static_model_preferences` with seeds, since it would require something into the initializer and I'd like to keep all the seed configuration in one place. We could create an initializer within the stripe installation but I'd like to keep this as simple as possible. What do you think?

lib/generators/solidus_stripe/install/install_generator.rb
kennyadsl: You are right, it works even with just calling the install with ENV vars. I think it will make the whole thing simpler, going to change it, thanks!

lib/tasks/solidus_stripe/db/seed.rake
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+namespace :db do
+  namespace :seed do
+    desc 'Loads stripe sample data'
+    task solidus_stripe: :environment do
+      seed_file = Dir[File.join(SolidusStripe::Engine.root, 'db', 'seeds.rb')][0]
```
tvdeyen:
```rb
SolidusStripe::Engine.root.join('db', 'seeds.rb')
```
should work as well

tvdeyen (approved):
This looks great. Just a minor nit

spaghetticode (approved):
👍 great addition, thank you @kennyadsl!

jacobherrington (approved):
Great change, 👍!

PR #7, opened 2019-04-24
Add Rubocop for linting

aitbw:
In an effort to improve tooling and to achieve a more consistent development environment, this PR aims to include Rubocop to this gem, as well as fix all the offenses based on the [Rubocop config file](https://github.com/solidusio/solidus/blob/master/.rubocop.yml) that Solidus uses.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,319 @@
+# Relaxed.Ruby.Style
+
+AllCops:
+  Exclude:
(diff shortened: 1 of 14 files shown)


tvdeyen (approved):
Thanks. We might want to enable HoundCI after this is merged.

jacobherrington (approved):
🐶

Issue #5, opened 2019-02-14
Better PCI-Compliance with stripe.js version 3


straumer:
As mentioned here:

Seems to do its job well with the solidus_backend and solidus_frontend. It'd also be nice if it had an option to use stripe.js version 3 out of the box, for better [PCI-Compliance](https://stripe.com/docs/security#validating-pci-compliance).

_Originally posted by @straumer in https://github.com/solidusio-contrib/solidus_stripe/issues/4#issuecomment-463804953_

hkrdrm:
How is solidus_stripe implemented out the box.  Is it using ruby backend to communicate with stripe.  I've got SSL certs setup with lets encrypt are there any additional steps I need to take to make sure I'm compliant.  I was reading something in stripe about uploading extra documentation to them if your not using the js version of stripe.

straumer:
For the better PCI-Compliance (not having to submit those extra docs), this gem needs an update for the solidus_frontend and solidus_backend OR you can make your own frontend that passes a payment token instead of credit card id to this payment method.

jackie577:
Hi, just wondering if anyone is working on this. I have been getting this error: "Sending credit card numbers directly to the Stripe API is generally unsafe." Anyone know what is the best way to handle? There is an option in Stripe to check "I agree that I am unsafe" but that doesn't feel the most satisfying.

spaghetticode:
[This PR](https://github.com/solidusio-contrib/solidus_stripe/pull/9) should allow addressing the issue.

spaghetticode:
Closing after #9 was merged.

PR #9, opened 2019-04-29
Add Stripe.js V3 with Elements

spaghetticode:
Ref #5

Besides the old Stripe.js V2 API, it's now possible to use the recent Stripe.js V3 library with Elements. 

A new model preference was added to `Spree::PaymentMethod::StripeCreditCard` allowing to choose
between the two different libraries in the admin area: 

Schermata 2019-04-29 alle 16 40 00

The new combination of Stripe.js V3 with Elements allows PCI SAQ-A certification eligibility.


jackie577:
Hi there, thanks for working on this! Would love to know if/when you are planning to release this. (:

spaghetticode:
@jackie577 now that specs are green I think the maintainers will review the code, hopefully it will get their approval and be merged soon.

lib/generators/solidus_stripe/install/install_generator.rb
```diff
@@ -3,6 +3,10 @@ module Generators
     class InstallGenerator < Rails::Generators::Base
       class_option :auto_run_migrations, type: :boolean, default: false
 
+      def add_stylesheets
+        inject_into_file "vendor/assets/stylesheets/spree/frontend/all.css", " *= require spree/frontend/solidus_stripe\n", before: '*/', verbose: true
```
jacobherrington: Is there a scenario where someone uses this library and `vendor/assets/stylesheets/spree/frontend/all.css` isn't present? I believe this will throw an error. Will this work if someone tries to use this library with a custom frontend?

jacobherrington:
Thanks @spaghetticode, this looks great. I have one question, but otherwise, I'm happy with the PR.

lib/generators/solidus_stripe/install/install_generator.rb
spaghetticode: @jacobherrington you're right. I tested it right now by changing the name of the file to a non-existing path, such as 

```inject_into_file "vendor/fake/frontend/fake.css", " *= require spree/frontend/solidus_stripe\n", before: '*/', verbose: true```

and Thor throws an error. 

What about adding an `if` statement that checks the existence of the file? I think this works fine: 

``` 
filename = 'vendor/assets/stylesheets/spree/frontend/all.css'
if File.file? filename
  inject_into_file filename, " *= require spree/frontend/solidus_stripe\n", before: '*/', verbose: true
end
```

I already added a commit to verify it works on Travis too.

Not sure I understand what you mean with the last question. The added styles are just cosmetics, and everything works fine also without them. Stripe.js V3 makes some assumptions regarding the HTML and the JS behavior, see their [tutorial](https://stripe.com/docs/stripe-js/elements/quickstart). As long as the custom frontend respects these assumptions then it's good to go.

jacobherrington (approved):
LGTM. Good work @spaghetticode!

PR #12, opened 2019-06-14
Use preferred_v3_elements instead of preferences[:v3_elements]

ChristianRimondi:
Using static model preferences, the line `preferences[:v3_elements]` throws a
```ruby
undefined method '[]' for #<Spree::Preferences::StaticModelPreferences::Definition:0x00007fb3c8a9b1a8>
```
so I had to change it to `preferred_v3_elements`.


spaghetticode (approved):
@ChristianRimondi thank you for the fix 👍

PR #13, opened 2019-07-10
Update README installation instructions

hashrocketeer:
It appears that this gem is not currently hosted on rubygems, so the installation instructions on the README should be updated so users know they should install from GitHub.

Related to this issue: https://github.com/solidusio-contrib/solidus_stripe/issues/4

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Installation
 In your Gemfile:
 
 ```ruby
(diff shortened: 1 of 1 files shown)


kennyadsl:
Thanks, I'll update the README if we decide to release on RubyGems!

kennyadsl (approved):

PR #17, opened 2019-09-25
Fix reusing cards with Stripe v3

### LINKED ISSUE #16: Stripe checkout fails when using a stored credit card number
We're using this gem to implement stripe checkout and it works fine the first time the user places an order. But on subsequent checkouts it fails if the user selects the "Use an existing card on file" option on the payments page. It looks to me like the gem expects the user to enter a credit card every time but maybe I'm misreading the code. We get this JS error in the console:

```
TypeError: element.data is not a function
```

```
enableElement (jquery_ujs.self-784a997f6726036b1993eb2217c9cb558e1cbb801c6da88105588c56f13b466a.js?body=1:383)
(anonymous) (payment:363)
setTimeout (async)
(anonymous) (payment:362)
Promise.then (async)
(anonymous) (payment:359)
dispatch (jquery3.self-e200ee796ef24add7054280d843a80de75392557bb4248241e870fac0914c0c1.js?body=1:5238)
elemData.handle (jquery3.self-e200ee796ef24add7054280d843a80de75392557bb4248241e870fac0914c0c1.js?body=1:5045)```

kennyadsl:
This PR fixes using existing cards with Stripe v3, which was not working (closes #16) and not tested (even v2 was not tested).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,7 +4,7 @@ AllCops:
   Exclude:
     - 'spec/dummy/**/*'
     - 'bin/**'
-  TargetRubyVersion: 2.2
+  TargetRubyVersion: 2.4
 
 # Sometimes I believe this reads better
 # This also causes spacing issues on multi-line fixes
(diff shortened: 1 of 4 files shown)


milesegan:
This seems to be working here. Thanks for the quick fix @kennyadsl .

spaghetticode (approved):
@kennyadsl thank you for this fix 👍

Issue #18, opened 2019-09-27
Handling of Strong Customer Authentication (SCA)


Aflier:
Using the the  Strong Customer Authentication (SCA) card (https://stripe.com/docs/payments/cards/testing) the payment cannot proceed. Payments from the EU countries will start failing. Thanks in advance.

spaghetticode:
This should be a good starting point: https://stripe.com/docs/payments/payment-intents, specifically:

> Anyone affected by SCA should migrate to the Payment Intents API to take advantage of exemption logic and a simpler integration. Integrations using the Charges API are not SCA-ready.

I also confirm that in tests, using 3D secure test cards such as `4000 0000 0000 3220` will fail the payment:

<img width="958" alt="Schermata 2019-09-27 alle 14 40 50" src="https://user-images.githubusercontent.com/141220/65769817-d59e9680-e134-11e9-9b5f-402b760c8491.png">

andyjeffries:
Without this change, it makes the gem pretty much unusable for anyone in the EU. Any updates? 

/cc: @kennyadsl (seems to be the main recent contributor)

kennyadsl:
We know this is missing but at the moment we are able to contribute with things that are needed by our customers at Nebulab and if they want to share solutions with the rest of the community, which is pretty common. The real issue here is that we only have US customers using Stripe at the moment so we had no need of adding support for SCA yet.

Solidus is an open-source project so everyone is welcome to submit a PR for that.

spaghetticode:
I'm working on this feature, but I can't put much time into it so it will take a while (I hope it will be done in one or two weeks).

kennyadsl:
@spaghetticode thanks! I just found [this post on testing Stripe SCA with Capybara](https://gorails.com/blog/fill-in-stripe-elements-js-for-sca-3d-secure-2-and-capybara) that maybe could be useful.

stuffmatic:
Hi! Any news on this?

kennyadsl:
Yes, we are planning some work on this soon.

leonjmac:
UK based dev here using this awesome gem. Have seen a few payment failures so far but not at a showstopper level. That said, if there's a branch where @spaghetticode, @kennyadsl, et al are working on implementing SCA compliance, please point me in the right direction as I'd like to help out... if I can!

kennyadsl:
[...]

PR #19, opened 2019-11-29
Add Stripe Payment Intents payment method for handling SCA transactions

spaghetticode:
Ref #18 

This is a WIP attempt in making this gem work with the Stripe Payment Intents API in order to be SCA compliant. Early sharing of the code should provide enough feedback to see if  this implementation makes sense and if the code actually works.

This PR is based on Stripe online documentation that illustrates how to integrate Payment Intents on top of Elements and other similar resources, and it required adding the `stripe` gem to the extension, as ActiveMerchant adapter for Payment Intents seems to follow a different flow.

This PR includes dedicated tests for the Payment Intents, and they seem to work seamlessly, but the code has not been tested on a real production Stripe account, so if anybody is interested in trying it on their responsibility then please let us know how it goes.

I'm also considering the effort to remove the `stripe` gem and use the ActiveMerchant Payment Intents adapter, but I think it won't be easy anyway.


kennyadsl:
@spaghetticode can we close this one in favor of #20?

spaghetticode:
Sure! Sorry I kept this open more time than what was necessary 😅

PR #20, opened 2019-12-13
Use Payment Intents API with Active Merchant

spaghetticode:
Stripe Payment Intents API must be used in order to handle SCA transactions. This PR is based on the official online Stripe documentation but uses the Payment Intents ActiveMerchant adapter instead of the `stripe` gem from Stripe guides. 

My first attempt to deliver SCA transactions was based on the `stripe` gem and is visible [here](https://github.com/solidusio/solidus_stripe/pull/19). This second approach has the merit to avoid adding a new gem to this extension just for handling the Payment Intents API.

The code is untested in production, so any feedback from people willing to use it there is welcome.


app/controllers/spree/stripe_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+            confirm: true,
+            setup_future_usage: 'on_session',
+            metadata: { order_id: current_order.id }
+          )
+        elsif params[:payment_intent_id].present?
+
+          intent = stripe.confirm_intent(params[:payment_intent_id], nil)
+        end
+      rescue Stripe::CardError => e
+        # Display error on client
```
kennyadsl: Maybe we can remove this line?

app/controllers/spree/stripe_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+        return
+      end
+
+      generate_payment_response(intent)
+    end
+
+    private
+
+    def stripe
+      @stripe ||= Spree::PaymentMethod::StripeCreditCard.first
```
kennyadsl: I'd love to find an alternative way for this. I saw the `payment_method_id` that we could use but maybe it's not always available?

app/models/spree/payment_method/stripe_credit_card.rb
```diff
@@ -62,20 +79,32 @@ def create_profile(payment)
...
-
+          if v3_intents?
+            payment.source.update_attributes!(
+              cc_type: payment.source.cc_type,
+              gateway_customer_profile_id: response.params['customer'],
+              gateway_payment_profile_id: response.params['id']
+            )
+          else
+            payment.source.update_attributes!(
+              cc_type: payment.source.cc_type, # side-effect of update_source!
```
kennyadsl: Can't really get the comment ` # side-effect of update_source!` here. Can you please expand a little bit?

solidus_stripe.gemspec
```diff
@@ -31,7 +31,7 @@ Gem::Specification.new do |s|
   # to the stripe gateway.
   #
   # This was resolved in v1.60, but we still need to skip 1.58 & 1.59.
-  s.add_dependency "activemerchant", "~> 1.48", "!= 1.58.0", "!= 1.59.0"
+  s.add_dependency "activemerchant", ">= 1.100" # includes "Stripe Payment Intents: Fix fallback for Store"
```
kennyadsl, line 37: I feel like the comment should be moved into the commit description, maybe with a specific commit?

kennyadsl (commented):
Great work @spaghetticode, I left some comments, let me know your thoughts!

app/controllers/spree/stripe_controller.rb
spaghetticode: Sure... comments are from the Stripe documentation example, I should have removed them in the first place 😅

app/models/spree/payment_method/stripe_credit_card.rb
spaghetticode: I don't know either, that comment was already there. If it has no significant meaning for you as well, probably it makes sense to just remove it.

app/controllers/spree/stripe_controller.rb
spaghetticode: We need to access that record in the controller because its class encapsulates all Stripe related knowledge, the previous one and the new one introduced with this PR. I can't think about a way to do things differently without changing much more code... do you have any suggestion?

kennyadsl: Does it need to be taken from the database? Just thinking aloud here, but maybe we can do `Spree::PaymentMethod::StripeCreditCard.new` and not relying on having the payment method saved? Just wondering if it can create issues if a store has two stripe payment methods on the db for some reason.

app/models/spree/payment_method/stripe_credit_card.rb
kennyadsl: I doesn't tell me anything, I'd remove it.

app/controllers/spree/stripe_controller.rb
spaghetticode: @kennyadsl with the help of @DanielePalombo I understood the issue you exposed. It's now fixed by loading the right payment method dynamically.

kennyadsl (approved):
Thanks @spaghetticode!

PR #21, opened 2020-01-10
Add solidus_dev_support gem

spaghetticode:
This PR fixes CI by introducing the gem `solidus_dev_support`


aldesantis:
@spaghetticode can you also run `solidus extension .` to see if any of the configuration files need to be updated? Note that we've literally just released solidus_dev_support 0.4.0 so you may want to `bundle update` first.

kennyadsl:
@spaghetticode do you have any idea about what happened to these pending specs? 🤔

app/assets/javascripts/spree/backend/solidus_stripe.js
```diff
@@ -0,0 +1,2 @@
+// Placeholder manifest file.
```
aldesantis: I think you can remove this and other empty asset files.

config/locales/en.yml
```diff
@@ -0,0 +1,5 @@
+# Sample localization file for English. Add more files in this directory for other locales.
```
aldesantis: This can be removed.

lib/generators/solidus_stripe/install/install_generator.rb
```diff
@@ -3,42 +3,28 @@
...
-        if options.migrate? && running_migrations?
+        run_migrations = options[:auto_run_migrations] || ['', 'y', 'Y'].include?(ask('Would you like to run the migrations now? [Y/n]'))
+        if run_migrations
           run 'bundle exec rake db:migrate'
         else
-          puts "Skiping rake db:migrate, don't forget to run it!"
-        end
-      end
-
-      def populate_seed_data
```
aldesantis: I think this logic is actually needed, right?

aldesantis (changes requested):

app/assets/javascripts/spree/backend/solidus_stripe.js
spaghetticode: sure. I was thinking that maybe those empty files may be handy in case somebody in the future needs to edit them. But surely they can add them back, why keep all that now-useless bolierplate around?

lib/generators/solidus_stripe/install/install_generator.rb
```diff
@@ -3,26 +3,28 @@
 module SolidusStripe
   module Generators
     class InstallGenerator < Rails::Generators::Base
-      class_option :migrate, type: :boolean, default: true
       class_option :auto_run_migrations, type: :boolean, default: false
-      class_option :auto_run_seeds, type: :boolean, default: false
+
+      def add_javascripts
```
kennyadsl: should we also remove these requires?

spaghetticode: I was thinking the same... but maybe it's better to conditionally load the files if present (and have `solidus_dev_support` do it). I think it makes sense to keep this kind of boilerplate. @aldesantis what do you think?

kennyadsl: I think this issue is related https://github.com/solidusio/solidus_dev_support/issues/45

aldesantis: We'll get to it! 🙏

aldesantis (approved):

aldesantis:
@spaghetticode looking good overall, but you'll need to address the failing specs.

Issue #22, opened 2020-01-20
Add support for Apple Pay through Stripe Payment Request Button


kennyadsl:
We need to add support for Apple Pay, with v3 Elements enabled, by adding support for [Payment Request Button](https://stripe.com/docs/stripe-js/elements/payment-request-button).

I did some experiment with @aldesantis today and to display the button it's just matter of adding the following code to the stripe v3 partial currently on master:

```diff
diff --git a/lib/views/frontend/spree/checkout/payment/v3/_stripe.html.erb b/lib/views/frontend/spree/checkout/payment/v3/_stripe.html.erb
index 2369d82..ce37cc7 100644
--- a/lib/views/frontend/spree/checkout/payment/v3/_stripe.html.erb
+++ b/lib/views/frontend/spree/checkout/payment/v3/_stripe.html.erb
@@ -1,3 +1,7 @@
+<div id="payment-request-button">
+  <!-- A Stripe Element will be inserted here. -->
+</div>
+
 <%= image_tag 'credit_cards/credit_card.gif', id: 'credit-card-image' %>
 <% param_prefix = "payment_source[#{payment_method.id}]" %>

@@ -62,14 +66,37 @@
     }
   };

+  var paymentRequest = stripe.paymentRequest({
+    country: 'US',
+    currency: 'usd',
+    total: {
+      label: 'Demo total',
+      amount: 1099,
+    },
+    requestPayerName: true,
+    requestPayerEmail: true,
+  });
+
   var cardNumber = elements.create('cardNumber', {style: style});
   var cardExpiry = elements.create('cardExpiry', {style: style});
   var cardCvc = elements.create('cardCvc', {style: style});
+  var prButton = elements.create('paymentRequestButton', {
+    paymentRequest: paymentRequest,
+  });

   cardNumber.mount('#card_number');
   cardExpiry.mount('#card_expiry');
   cardCvc.mount('#card_cvc');

+  // Check the availability of the Payment Request API first.
+  paymentRequest.canMakePayment().then(function(result) {
+    if (result) {
+      prButton.mount('#payment-request-button');
+    } else {
+      document.getElementById('payment-request-button').style.display = 'none';
+    }
+  });
+
   $(function() {
     var form = Spree.stripePaymentMethod.parents('form');
     var submitButton = form.find('input[type="submit"]');
```

Of course, it needs to have dynamic data taken from the current order. 

[...]

PR #25, opened 2020-01-30
Allow Apple Pay and Google Pay via Payment Request Button

spaghetticode:
Ref #22 

Payments with ApplePay and GooglePlay can be managed with Stripe only by using the [Payment Request Button API](https://stripe.com/docs/stripe-js/elements/payment-request-button) which integrates nicely with the recently added [Payment Intents](https://github.com/solidusio/solidus_stripe/pull/20) API. 

One notable concern is that this feature cannot be easily tested with automated tests, as it requires that:
* the server runs on HTTPS with a valid certificate and is accessible from the outer Internet (Stripe verifies the validity of the setup before allowing the use of the button)
* with ApplePay, a custom file with a verification code must be downloaded from the Stripe account and made available from the app to the outer Internet (again, Stripe will verify the file contents match before enabling the functionality)

Due to these hard-to-fulfill requirements for a testing environment no automated tests are provided. Please consider that a part of the functionality is based on the Payment Intents API, which is on the other hand well tested, as it does not have such strong requirements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ Spree.config do |config|
     'stripe_env_credentials',
     secret_key: ENV['STRIPE_SECRET_KEY'],
     publishable_key: ENV['STRIPE_PUBLISHABLE_KEY'],
+    stripe_country: 'US',
     v3_elements: false,
     v3_intents: false,
     server: Rails.env.production? ? 'production' : 'test',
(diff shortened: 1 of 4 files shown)


kennyadsl (approved):
Thanks @spaghetticode, this is awesome!

kennyadsl:
Just one thing: the correct names are `Apple Pay` and `Google Pay`, with a space in between. 🤓

PR #28, opened 2020-02-06
Remove ERB from Elements and Intents JS code

spaghetticode:
In order to be able to move the JS business logic for Stripe payments with Intents and Elements API to an external JS file, we need to get rid of the ERB interpolation used to pass Stripe and payment config values from Ruby to the JS.


kennyadsl (approved):
Thanks @spaghetticode

PR #29, opened 2020-02-13
Cart page payment with Stripe payment request button (Apple/Google Pay) 

spaghetticode:
This PR allows to pay with Apple Pay and Google Pay directly from the cart page, thus skipping much of the standard Solidus checkout process. 

Store developers must select the payment method to be used by overriding the helper method `Spree::OrdersHelper#cart_checkout_payment_method`. 

Setting up an integration test environment capable of fulfilling Stripe rules (externally reachable domain with valid https certificate, verification file for Apple Pay) for the payment request button is quite hard, so the functionality comes without automated integration tests. Tests exist only for ruby service objects used during the payment process. 

Also, an extensive rewrite of the JS code involved in Stripe Elements and Payment Intents processes was performed in order to remove much duplication and try to simplify the flow.


README.md
```diff
@@ -106,6 +106,16 @@ Check the Payment Intents section for setup details. Also, please
...
+It's possible to pay with Apple Pay and Google Pay directly from the cart
+page. In order to enable the functionality, you have to override the
+view helper method `Spree::OrdersHelper#cart_checkout_payment_method` and
+make it return the `Spree::PaymentMethod::StripeCreditCard` instance that
+you want to use for the checkout. Of course, rules stated in the paragraph
+above must apply also for this payment method.
+
+Once the helper method is set, a view partial with the relevant HTML and
+JS code will be injected via `deface` in the cart page.
+
```
kennyadsl, line 125: What about just providing a partial with instructions without using deface at all?

app/assets/javascripts/solidus_stripe/stripe-init.js
```diff
@@ -0,0 +1,173 @@
+Spree.stripePaymentMethod = {
```
kennyadsl: I think this should just be a maniefest, and should require a file with this content. This way it should be easier for stores to override it with a custom version.

app/controllers/solidus_stripe/intents_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+            confirmation_method: 'manual',
+            confirm: true,
+            setup_future_usage: 'on_session',
+            metadata: { order_id: current_order.id }
+          )
+        elsif params[:stripe_payment_intent_id].present?
+          intent = stripe.confirm_intent(params[:stripe_payment_intent_id], nil)
+        end
+      rescue Stripe::CardError => e
+        render json: { error: e.message }
```
kennyadsl: We should add the correct status code to this render, by default it's `200` even if they are errors. Something like:

```ruby
 render json: { error: e.message }, status: 500
```

(same in the other json errors below)

app/decorators/controllers/frontend/orders_controller/cart_checkout_payment_method_decorator.rb
```diff
@@ -0,0 +1,17 @@
...
+  def self.prepended(base)
+    base.helper_method :cart_checkout_payment_method
+  end
+
+  # Override this method in order to enable cart page checkout
+  # with Apple Pay and Google Pay. This method must return the
+  # `Spree::PaymentMethod::StripeCreditCard` instance that will
+  # be used for configuring the payment request button, for
+  # example `Spree::PaymentMethod::StripeCreditCard.first`
+  def cart_checkout_payment_method
```
kennyadsl: I think we might try to remove this file. If we remove deface, we leave to users the responsibility to get this method by suggesting them to use the partial:

```erb
<%= render 'stripe_payment_request_button', cart_checkout_payment_method: Spree::PaymentMethod::StripeCreditCard.first %>
```

This way, it will be explicit that it's up to them to pick the right payment method.

kennyadsl (commented):
Left some initial discussion points, let me know your thoughts.

lib/views/frontend/spree/orders/_stripe_payment_request_button.html.erb
```diff
@@ -0,0 +1,48 @@
+<% if current_order.present? && cart_checkout_payment_method.present? %>
+  <div id="payment-request">
+    <%= form_tag update_checkout_path(:payment), method: :patch do %>
```
kennyadsl: Is this form really needed? I think it could be better to fully handle this via JS since this partial could be added inside another form and this could lead to issues.

app/controllers/solidus_stripe/intents_controller.rb
kennyadsl: Still missing here?

lib/views/frontend/spree/orders/_stripe_payment_request_button.html.erb
```diff
@@ -0,0 +1,48 @@
...
+      <div id="payment_method_<%= cart_checkout_payment_method.id %>">
+        <%= hidden_field_tag "order[payments_attributes][][payment_method_id]", cart_checkout_payment_method.id %>
+
+        <div id="payment-request-button" data-stripe-config="<%= cart_checkout_payment_method.stripe_config(current_order).to_json %>"></div>
```
kennyadsl: Can we also add some HTML class to this element?

lib/views/frontend/spree/orders/_stripe_payment_request_button.html.erb
```diff
@@ -0,0 +1,48 @@
+<% if current_order.present? && cart_checkout_payment_method.present? %>
+  <div id="payment-request">
```
kennyadsl: can we add an HTML class here as well? it probably needs some style in the host application.

kennyadsl (commented):
A second round of comments after trying it on a real store.

lib/views/frontend/spree/orders/_stripe_payment_request_button.html.erb
spaghetticode: It would imply modifying some JS code. At the moment, as much JS as possible is shared between the Elements, Payment Intents, and cart checkout button functionalities.

The first two implementations use the form in order to collect customer's data and submit payment information to the application, in this case this is not strictly necessary as the browser is doing it, but by using a form that behaves like the other twos we get to retain the same behavior and reuse the JS code. 

I can try to remove it so we can compare the results, what do you think?

kennyadsl: If we can I would give it a try since nested forms are not allowed for valid HTML.

app/assets/javascripts/solidus_stripe/stripe-init/base.js
```diff
@@ -0,0 +1,173 @@
...
+      requestShipping: config.requestShipping,
+      shippingOptions: [
+      ]
+    });
+
+    var prButton = elements.create('paymentRequestButton', {
+      paymentRequest: paymentRequest
+    });
+
+    paymentRequest.canMakePayment().then(function(result) {
```
kennyadsl, line 74: I'd go with this version that hides the main wrapper if there's no payment button available.

```js
    document.getElementById('stripe-payment-request').style.display = 'none';
    paymentRequest.canMakePayment().then(function(result) {
      if (result) {
        document.getElementById('stripe-payment-request').style.display = '';
        prButton.mount('#payment-request-button');
      } else {
        document.getElementById('stripe-payment-request').style.display = 'none';
      }
    });
```

kennyadsl (approved):

PR #31, opened 2020-03-03
Refactor Stripe V3 Intents, Elements and cart checkout JS code

spaghetticode:
The code has been encapsulated in proper constructor objects by separating common code as necessary.

Code shared among all the Stripe V3 API checkout procedures is now contained in the `Spree.StripePayment` constructor.

`SolidusStripe.Elements` inherits directly from `SolidusStripe.Payment`, same goes for `SolidusStripe.CartPageCheckout`.

`SolidusStripe.PaymentIntents` inherits from `SolidusStripe.Elements`, as it basically extends the Elements functionality by adding the 3D secure feature.

Both `SolidusStripe.CartPageCheckout` and `SolidusStripe.PaymentIntents` allow to pay with Apple Pay/Google Pay, so they include shared methods from `PaymentRequestButtonShared`.

All functions are defined on the different classes prototype, this will allow to easily customize those functions according to needs by just overriding them on the proper prototype object.


kennyadsl (approved):
Thanks, @spaghetticode!

PR #32, opened 2020-03-05
Manage Stripe V3 JS code via Sprokets

spaghetticode:
Stripe V3 JS code (Elements, Payment Intents, and cart checkout button)
has been completely removed from HTML views and is now loaded via
Sprockets in Solidus `frontend/all.js` file.

The right API (Elements, Intents...) is now selected by looking for
an HTML element with `data-v3-api` attribute set with the name of
one of Stripe V3 APIs.


lib/views/frontend/spree/checkout/payment/v3/_form_elements.html.erb
```diff
@@ -1,4 +1,4 @@
-<div id="payment-request-button" data-stripe-config="<%= payment_method.stripe_config(current_order).to_json %>"></div>
+<div id="payment-request-button" data-stripe-config="<%= payment_method.stripe_config(current_order).to_json %>" data-v3-api="<%= stripe_v3_api %>"></div>
```
kennyadsl, line 1: Do we always show this button in v3, or it should be wrapped in an if somehow? Or is it already done via JS?

kennyadsl (commented):
Love this. 

I left a question and I also have some doubt about how to customize some part of the provided JS now that it is in the assets pipeline. I think it's pretty easy but maybe adding a couple of lines in the README will help new users. What do you think?

lib/views/frontend/spree/checkout/payment/v3/_form_elements.html.erb
spaghetticode: Currently, there's one flag to disable the button: 

```ruby
def payment_request?
  v3_intents? && preferred_stripe_country.present?
end
```
when this method returns false, the button is not shown. So, the preference `stripe_country` controls the button visibility, as it's a requirement for the payment request to work.

When the preference is not present, the JS code eventually will not initialize the button:

```js
    setUpPaymentRequest: function(opts) {
      var opts = opts || {};
      var config = this.config.payment_request;

      if (config) {
       // ....
```

It's probably worth documenting a bit this as well.

spaghetticode:
@kennyadsl I agree adding some documentation will help, will do that soon, thanks for the suggestion!

kennyadsl (approved):
👍

PR #35, opened 2020-03-13
Change order description that we pass to Stripe

kennyadsl:
It's time to change the description that we pass to Stripe when creating a new order.


spaghetticode (approved):
@kennyadsl thank you 👍

PR #37, opened 2020-03-17
Add hook to provide custom Stripe Elements options

stuffmatic:
The Stripe Elements instance is currently created with the hardcoded options `{ locale: 'en'}`. This pull request adds a hook similar to `baseStyle` for providing custom options, for example a different locale or custom fonts. The supported options are documented [here](https://stripe.com/docs/js/elements_object/create).

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,12 @@ SolidusStripe.CartPageCheckout.prototype.onPrButtonMounted = function(id, result
 }
 ```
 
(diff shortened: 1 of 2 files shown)


spaghetticode (approved):
@stuffmatic thank you for the improvement 👍

kennyadsl (approved):
Thanks, makes perfect sense.

Issue #36, opened 2020-03-17
Visa credit card type is blank


stuffmatic:
I just noticed the [mapCC](https://github.com/solidusio/solidus_stripe/blob/fc5cc7fd8518acbed2d7d7dc91def6abe6698614/app/assets/javascripts/spree/frontend/solidus_stripe/stripe-elements.js#L89) function seems to expect the string `Visa` for Visa cards, whereas the Stripe API seems to return `visa`, which causes the `cc_type` of the `Spree::CreditCard` to be empty. 

The example response in [this section of the Stripe API](https://stripe.com/docs/api/metadata) has `brand` set to `visa`, which is what I'm seeing. Intrestingly, [this section](https://stripe.com/docs/api/cards/object#card_object-brand) seems to suggest it should be `Visa`.

I'm using the test Visa card 4242 4242 4242 4242 and have `v3_intents` set to `true`. I haven't yet been able to test this for other card types, so I can't suggest a general solution at this point.

Finally, a huge thanks for your work on this gem!

spaghetticode:
`mapCC` is a rather old function, so it's probably not up-to-date with all Stripe library changes. 
I think that problem can be solved by converting the equality checks on strings to checks on regexp, @stuffmatic what do you think?

stuffmatic:
@spaghetticode I haven't been able to find any clear documentation on this yet. However, I found a clue in the official Android SDK, where [CardBrand.tk](https://github.com/stripe/stripe-android/blob/master/stripe/src/main/java/com/stripe/android/model/CardBrand.kt) is an enum class describing credit card brands. The members have both `displayName` and `code` properties, where the latter seems to correspond to the lower case `visa` I saw.

Here are the `displayName` - `code` pairs from CardBrand.tk

* American Express - amex
* Diners Club - diners
* Discover - discover
* JCB - jcb
* Mastercard - mastercard
* Unionpay - unionpay
* Visa - visa
* Unknown - unknown

The list above covers all valid card brands given [here](https://stripe.com/docs/api/cards/object).

Also, I tried some [test cards](https://stripe.com/docs/testing#cards) of different brands, and the ones I was allowed to use resulted in a `brand` value matching the lower case values in the list above.

So it looks like the safest fix would be to do double string comparisons in `mapCC`, so instead of

`
if (ccType === 'American Express')
`

do 

`
if (ccType === 'American Express' || ccType === 'amex')
`

Edit: I created a PR with my suggested fix.

[...]

PR #38, opened 2020-03-18
Fix stripe-to-solidus card type mapping

stuffmatic:
The Stripe API now seems to describe credit card brands by their code (e.g `amex`) as opposed to their display name (e.g `American Express`), which results in a blank `cc_type` for `Spree::CreditCard`. 

This pull request attempts to fix this by making the `mapCC` function  compare against both the brand code and the brand display name.


kennyadsl (approved):
Thanks! Ref #36

spaghetticode (approved):
@stuffmatic thanks for this, it's much appreciated 👍

PR #43, opened 2020-03-27
Payment Intents with card pre-authorization

spaghetticode:
At the moment payments done with Payment Intents end up in a double charge for the customer. 

A first charge is created instantly when the customer authorizes their card, and a second one is created later after order confirmation. This issue is not of easy solution, as the first charge is always required for Payment Intents, and the second one is created automatically by Solidus flow.
Also, the first charge happens outside of Solidus' current processes, so there's no record of it in on the DB.

A possible temporary workaround for this issue is to introduce the concept of card pre-authorization: the first charge, now limited to only 1$ is not captured anymore, and have an appropriate description so, even when customers receive notifications for uncaptured transactions from their bank, they will understand this is not going to be actually captured.

### DIFF
--- a/app/controllers/solidus_stripe/intents_controller.rb
+++ b/app/controllers/solidus_stripe/intents_controller.rb
@@ -8,12 +8,14 @@ module SolidusStripe
       begin
         if params[:stripe_payment_method_id].present?
           intent = stripe.create_intent(
-            (current_order.total * 100).to_i,
+            100,
             params[:stripe_payment_method_id],
+            description: "Card pre-authorization for order #{current_order.number}",
             currency: current_order.currency,
             confirmation_method: 'manual',
+            capture_method: 'manual',
             confirm: true,
-            setup_future_usage: 'on_session',
+            setup_future_usage: 'off_session',
             metadata: { order_id: current_order.id }
           )
         elsif params[:stripe_payment_intent_id].present?
@@ -42,7 +44,7 @@ module SolidusStripe
             requires_action: true,
             stripe_payment_intent_client_secret: response['client_secret']
           }
-      elsif response['status'] == 'succeeded'
+      elsif %w[requires_capture succeeded].include?(response['status'])
         render json: { success: true }
       else
         render json: { error: response['error']['message'] }, status: 500
--- a/app/models/spree/payment_method/stripe_credit_card.rb
+++ b/app/models/spree/payment_method/stripe_credit_card.rb
@@ -142,7 +142,7 @@ module Spree
         options = {}
         options[:description] = "Solidus Order ID: #{transaction_options[:order_id]}"
         options[:currency] = transaction_options[:currency]
-
+        options[:off_session] = true if v3_intents?
         if customer = creditcard.gateway_customer_profile_id
           options[:customer] = customer
         end
--- a/README.md
+++ b/README.md
@@ -72,6 +72,15 @@ Stripe payments.
 Using Stripe Payment Intents API
 --------------------------------
 
+⚠️ The current implementation for the Payment Intents API in the gem creates
+double charges. This happens only when the customer adds a new credit card.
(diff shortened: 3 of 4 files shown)


app/decorators/solidus_stripe/active_merchant/stripe_payment_intents_gateway_decorator.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module ActiveMerchant::StripePaymentIntentsGatewayDecorator
+  def authorize(money, payment_method, options = {})
```
kennyadsl: can't we pass this option to AM via the authorize method in our StripePaymentMethod?

app/controllers/solidus_stripe/intents_controller.rb
```diff
@@ -8,12 +8,14 @@ def confirm
       begin
         if params[:stripe_payment_method_id].present?
           intent = stripe.create_intent(
-            (current_order.total * 100).to_i,
+            100,
```
kennyadsl, line 11: what's wrong with preauthorizing the full amount?

kennyadsl (commented):
Left a couple of questions, thanks Andrea!

aldesantis (approved):
@spaghetticode this is looking good as a workaround, although it's not an ideal UX. For a long-term solution, would it make sense to implement some kind of API in the core to support this?

spaghetticode:
@aldesantis the long term solution would be, in my opinion, to only save the Payment Intents charge in Solidus (with the full amount, uncaptured) and don't create the second one (which is currently created automatically). So, not sure what changes we would need inside Solidus, if any at all... this is something I still have to look into.

app/controllers/solidus_stripe/intents_controller.rb
spaghetticode: I think that would block twice the full amount on the customer card. 

Also, as a customer, I would not be particularly scared by noticing a $1 pre-authorization charge notice (I've seen something similar done in the past by PayPal and Amazon as well if I recall correctly), but seeing the double amount I would instantly think there's been some double charge error.

app/decorators/solidus_stripe/active_merchant/stripe_payment_intents_gateway_decorator.rb
spaghetticode: I think that makes more sense, thanks!

app/controllers/solidus_stripe/intents_controller.rb
kennyadsl: yep, but if you have $1 on your card, it's not guaranteed that you have the full amount to pay the order. Anyway, I'm ok with this strategy, and we can revisit it later. We just need to be clear in the README that this is the behavior of the gem. Also, what about pre-authorizing $0.01 instead?

spaghetticode: Unfortunately, you can't pre-authorize less than 0.50$ (or 0.50€... so why not 0.50? if you're in EU and you're buying in USD, the conversion would be 0.43€ which is not acceptable again).
I'm not sure what happens when you don't have enough funds for the order, but I guess that at that point the second payment would fail (for sure the capture would fail). 

Anyway, the second payment is Solidus' default process, for example what happens with Solidus Stripe and Elements. The failed pre-authorization for the whole amount would behave just like in that case.

app/models/spree/payment_method/stripe_credit_card.rb
```diff
@@ -72,6 +72,7 @@ def purchase(money, creditcard, transaction_options)
       end
 
       def authorize(money, creditcard, transaction_options)
+        transaction_options.merge!(off_session: true)
```
kennyadsl: do we need to add this option only when using intents?

spaghetticode:
Superseded by #45

PR #45, opened 2020-04-09
Create a single charge when using Stripe Payment Intents

### LINKED ISSUE #44: Duplicates charges with Payment Intents
At the moment the Payment Intents implementation creates duplicate charges for each order. This is visible only on the Stripe dashboard page. 

The first payment, currently invisible on Solidus (nothing is recorded on the DB) is created and charged instantly as soon as the customer authorizes their credit card. The second one is created after the order is finalized on Solidus backend, like it does for Elements and other payment implementations. This second charge needs to be captured later.

A possible workaround for this issue is proposed in #43, while waiting for a better solution to be prepared.

spaghetticode:
Fixes https://github.com/solidusio/solidus_stripe/issues/44, plus other minor issues found while working on that main problem.

When using the Payment Intents API, we need to create a single charge. In order to achieve the goal, we now create a Solidus payment as soon as the Payment Intent is authorized from Stripe (while the customer is still on the payment page). Later, we skip the creation of the second payment during Solidus default flow after the order has been confirmed. 

The first payment will be created `uncaptured` instead of `authorized` (already captured) as the customer may decide to not complete their purchase. 

When the customer enters multiple credit cards during the same checkout process or changes the payment method from a freshly entered credit card to something else (ie. check) we also need to properly invalidate (void) stale Payment Intents charges and their payment records in Solidus.

This PR also fixes an issue with 3D-Secure card reusability. before this fix, orders finalized by reusing an already authorized 3DS card would end up in creating payments that cannot be properly captured. 

Some integration tests were added in order to improve edge cases coverage, including checks on the backend that verify payments are created with the proper state and their capturability, when appropriate.

### DIFF
--- a/app/assets/javascripts/spree/frontend/solidus_stripe/stripe-payment-intents.js
+++ b/app/assets/javascripts/spree/frontend/solidus_stripe/stripe-payment-intents.js
@@ -69,6 +69,7 @@ SolidusStripe.PaymentIntents.prototype.onIntentsPayment = function(payment) {
         'Content-Type': 'application/json'
       },
       body: JSON.stringify({
+        form_data: this.form.serialize(),
         spree_payment_method_id: this.config.id,
         stripe_payment_method_id: payment.paymentMethod.id,
         authenticity_token: this.authToken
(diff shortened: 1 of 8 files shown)


kennyadsl (approved):
Thanks @spaghetticode, commits are super-readable and they made it easier to review.

PR #48, opened 2020-04-17
Relax solidus_support dependency

kennyadsl:
It allows to use the latest version of Solidus Support and also removes a deprecated module inclusion.


spaghetticode (approved):
@kennyadsl thank you!

PR #57, opened 2020-05-15
Fix order cancel with Payment Intents captured payment

### LINKED ISSUE #56: [ADMIN] Order cancel doen't work with Payment Intents captured payments
When trying to cancel an order with an already captured payment made using the Payment Intents API, the application raises an error with the following message from Stripe:

```
A PaymentIntent object can be canceled when it is in one of these statuses: 
requires_payment_method, requires_capture, requires_confirmation, requires_action
```
This error can be avoided by refunding the payment (instead of canceling it) during the order canceling process.

spaghetticode:
Fixes https://github.com/solidusio/solidus_stripe/issues/56

This PR adds the `Spree::PaymentMethod::StripeCreditCard#try_void` method that is now called when an order is canceled in the admin area (Solidus used to call the now deprecated method `Spree::PaymentMethod#cancel`).

If we're dealing with a Payment Intents payment, and the payment is already captured, Stripe will not allow canceling it, we can only refund it. 

Also, a custom refund reason record is added via migration, in order to be able to clearly determine those payments that were refunded because there was no alternative.

### DIFF
--- a/app/models/spree/payment_method/stripe_credit_card.rb
+++ b/app/models/spree/payment_method/stripe_credit_card.rb
@@ -81,6 +81,21 @@ module Spree
         gateway.void(response_code, {})
       end
 
+      def payment_intents_refund_reason
+        Spree::RefundReason.where(name: Spree::Payment::Cancellation::DEFAULT_REASON).first_or_create
+      end
+
+      def try_void(payment)
+        if v3_intents? && payment.completed?
+          payment.refunds.create!(
+            amount: payment.credit_allowed,
+            reason: payment_intents_refund_reason
+          ).response
+        else
+          payment.void_transaction!
+        end
+      end
+
       def cancel(response_code)
         gateway.void(response_code, {})
       end
--- a/app/decorators/models/spree/refund_decorator.rb
+++ b/app/decorators/models/spree/refund_decorator.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+module Spree
+  module RefundDecorator
+    attr_reader :response
+
+    Spree::Refund.prepend(self)
+  end
+end
--- a/spec/features/stripe_checkout_spec.rb
+++ b/spec/features/stripe_checkout_spec.rb
@@ -400,8 +400,9 @@ RSpec.describe "Stripe checkout", type: :feature do
         click_button "Place Order"
         expect(page).to have_content("Your order has been processed successfully")
 
-        # Capture in backend
         Spree::Order.complete.each do |order|
+          # Capture in backend
+
           visit spree.admin_path
 
           expect(page).to have_selector("#listing_orders tbody tr", count: 2)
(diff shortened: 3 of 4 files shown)


app/models/spree/payment_method/stripe_credit_card.rb
```diff
@@ -81,6 +83,19 @@ def void(response_code, _creditcard, _transaction_options)
         gateway.void(response_code, {})
       end
 
+      def try_void(payment)
+        if payment.completed?
+          reason = Spree::RefundReason.find_by(name: self.class::PAYMENT_INTENTS_REFUND_REASON) || Spree::RefundReason.first
```
ccarruitero: I think is just necessary `PAYMENTS_INTENTS_REFUND_REASON`, since the constant is in the same class.

Also, instead to fallback to first reason maybe we can use `find_or_create`?

spaghetticode: @ccarruitero thanks for spotting the issue with the constant, what was  I thinking when I wrote that?

I thought about using `find_or_create`, but then I went for the migration (the record is explicitly created in a more "expected" place) and the fallback to `Spree::RefundReason.first` for making the custom refund reason optional. 
But I think that your options are good as well. I'm wondering if a rake task for adding the refund reason might be a better option... does anybody want to share opinions on this?

kennyadsl: I have some concerns about creating a migration for this RefundReason. Can't we use an already existing one as default, like `Spree::Payment::Cancellation::DEFAULT_REASON`? Maybe also defined in a separate method so that people can customize it and use something else if needed?

app/models/spree/payment_method/stripe_credit_card.rb
```diff
@@ -81,6 +81,22 @@ def void(response_code, _creditcard, _transaction_options)
...
 
+      def payment_intents_refund_reason
+        Spree::RefundReason.where(name: Spree::Payment::Cancellation::DEFAULT_REASON).first_or_create
+      end
+
+      def try_void(payment)
+        if v3_intents? && payment.completed?
+          Spree::Refund.create!(
+            payment: payment,
+            amount: payment.amount,
```
rainerdema: I think you can use [credit_allowed](https://github.com/solidusio/solidus/blob/475d9db5d0291dd4aeddc58ec919988c336729bb/core/app/models/spree/payment.rb#L143-L145) here (like in [Spree::Payment::Cancellation#cancel](https://github.com/solidusio/solidus/blob/b8296bdb5c9651e60239ab27c0093636981fd0aa/core/app/models/spree/payment/cancellation.rb#L34))
I don't think it's a problem for Stripe but, if there are some partial refunds already done, `credit_allowed` should give the correct amount to be refunded. WDYT?

app/models/spree/payment_method/stripe_credit_card.rb
```diff
@@ -81,6 +81,21 @@ def void(response_code, _creditcard, _transaction_options)
         gateway.void(response_code, {})
       end
 
+      def payment_intents_refund_reason
+        Spree::RefundReason.where(name: Spree::Payment::Cancellation::DEFAULT_REASON).first_or_create
+      end
+
+      def try_void(payment)
+        if v3_intents? && payment.completed?
```
ChristianRimondi, line 89: Why don't you just return `false` here? It seems like solidus already handles the refund [here](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/payment/cancellation.rb#L30-L36) 🤔

spaghetticode: @ChristianRimondi that way the payment would remain to its original state, instead of becoming`void`, as this [line](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/payment/cancellation.rb#L32) would not run. That line executes [this method](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/payment/processing.rb#L190).

kennyadsl (approved):

PR #61, opened 2020-05-29
Add stripe name to payment form fields id

aleph1ow:
When the gem is installed into solidus, the default payment gateway needs to be
disabled to make the solidus_stripe js to work. This is due to ids conflict.
This commit differentiate the stripe's payment form fields ids so it can cohexist
with the default payment gateway.


spaghetticode:
@aleph1ow thanks for this fix, I think it's a good one! 

Specs are failing, I think that element selectors there must be updated according to the changes. Can you take a look? thanks 🙏

PR #66, opened 2020-06-17
Remove server and test_mode from README

adammathys:
As it turns out, neither of these values are used by any of the Stripe gateway code. They aren't referenced anywhere in this extension nor in either of the `ActiveMerchant` gateways. [1,2] Including them only serves to confuse users. (I know I was certainly confused when I came across some configuration that had them set to use "production" when they were instead connecting to a sandbox account.)

1. https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe.rb
2. https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe_payment_intents.rb

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,9 +56,7 @@ Spree.config do |config|
     publishable_key: ENV['STRIPE_PUBLISHABLE_KEY'],
     stripe_country: 'US',
     v3_elements: false,
-    v3_intents: false,
-    server: Rails.env.production? ? 'production' : 'test',
-    test_mode: !Rails.env.production?
+    v3_intents: false
   )
 end
 ```
@@ -100,9 +98,7 @@ Spree.config do |config|
     publishable_key: ENV['STRIPE_PUBLISHABLE_KEY'],
     stripe_country: 'US',
     v3_elements: false,
-    v3_intents: true,
-    server: Rails.env.production? ? 'production' : 'test',
-    test_mode: !Rails.env.production?
+    v3_intents: true
   )
 end
 ```


spaghetticode:
@adammathys I think they're used by Solidus in the [gateway](https://github.com/solidusio/solidus/blob/17c8cf5beaba89aabff93661cba85e6c77df5137/core/app/models/spree/payment_method.rb#L125) method and passed down to Active Merchant via the `test` option that is used in te hStripeGateway [superclass](https://github.com/activemerchant/active_merchant/blob/450ba97194ea6410dd40d9fd8d2280aa2a6ea53b/lib/active_merchant/billing/gateway.rb#L185).

adammathys:
@spaghetticode That is correct, but `Gateway#test?` isn't used by any of the code. It's not exposed by this extension or Solidus. And neither of the Stripe gateways use it.

Right now you can set both values to indicate that you want to be in production and have it still connect to a sandbox environment. Or possibly more worryingly, set them as if you're running against a test account and have all of your requests go to a production one instead. There are thankfully other safeguards in place on Stripe's end to protect you from doing anything too damaging in those scenarios, but I'd still maintain that the settings serve no real purpose for this extension other than to possibly confuse users.

aldesantis:
I agree that, even though the values are passed to ActiveMerchant, we shouldn't give people a false impression that they're actually having any effect. At some point, we'll have to completely rethink those two options by consolidating them into one and only showing them when it makes sense, but that will have to happen at the core level. For the time being, I'm happy to remove them from the readme.

spaghetticode (approved):

spaghetticode:
@adammathys can you please rebase with master? CI was failing due to some Stripe UI changes that are now addressed (at least until they change something else again 😭) . Thanks 🙏

PR #69, opened 2020-06-26
Save correct cardholder name in Spree::CreditCard

### LINKED ISSUE #68: Card name ignored when adding new card to an order during checkout
https://github.com/solidusio/solidus_stripe/blob/b9c0b8a1db057923176635f0a6d7b30c6c57f5c2/app/models/solidus_stripe/create_intents_payment_service.rb#L64

Currently when a user checks out and adds a new card the name they entire is ignored in favor of the name on the order billing address. Seems like this line should be using the name from the card and not the address, right?

spaghetticode:
When paying with Payment Intents either using the HTML form or the payment request button, we need to properly detect the cardholder name in order to save the correct information in the credit card DB record.

Fixes https://github.com/solidusio/solidus_stripe/issues/68

### DIFF
--- a/app/models/solidus_stripe/create_intents_payment_service.rb
+++ b/app/models/solidus_stripe/create_intents_payment_service.rb
@@ -61,7 +61,7 @@ module SolidusStripe
             cc_type: intent_card['brand'],
             last_digits: intent_card['last4'],
             gateway_payment_profile_id: intent_customer_profile,
-            name: address_full_name,
+            name: card_holder_name || address_full_name,
             address_attributes: address_attributes
           }
         }]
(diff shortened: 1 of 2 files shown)


aldesantis (approved):
@spaghetticode the change itself looks good, but with all these methods and conditionals that extract data from Stripe's response I feel like there's a lot of indirection going on which makes it hard to figure out what's happening. At some point (or even in this PR if you have time) we may want to streamline them by moving more stuff to `payment_params` — it may make the method longer, but it will also make it easier to understand.

aldesantis:
Sounds good, this is GTM as far as I'm concerned!

PR #75, opened 2020-07-15
Fix StripeCreditCard#try_void with Elements/V2 API

spaghetticode:
`Payment::Cancellation#cancel` calls StripeCreditCard#try_void`.

When not using Payment Intents (ie. when using Elements or V2 API), after calling `try_void`, the former makes the following call:

    payment.send(:handle_void_response, response)

which is already called by `payment.void_transaction!`, making the two incompatible, as `#handle_void_response` cannot be called twice here - it would raise the error:

    NoMethodError: undefined method `success?' for true:TrueClass

Under the hood `#void_transaction!` calls `payment_method.void`, so the new code is rather equivalent to the previous one. The only relevant feature we're losing is that `#void_transaction!` wraps the call in a `protect_from_connection_error` block.

The modified integration spec verifies that an order created with Elements can be canceled without raising the error mentioned above.

### DIFF
--- a/app/models/spree/payment_method/stripe_credit_card.rb
+++ b/app/models/spree/payment_method/stripe_credit_card.rb
@@ -92,7 +92,7 @@ module Spree
             reason: payment_intents_refund_reason
           ).response
         else
-          payment.void_transaction!
+          void(payment.response_code, nil, nil)
         end
       end
 
--- a/spec/features/stripe_checkout_spec.rb
+++ b/spec/features/stripe_checkout_spec.rb
@@ -240,54 +240,89 @@ RSpec.describe "Stripe checkout", type: :feature do
       expect(page).to have_content("Your order has been processed successfully")
     end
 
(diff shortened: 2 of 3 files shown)


aldesantis:
@spaghetticode do you think there is a way to maintain the `protect_from_connection_error` behavior? That seems quite useful.

spaghetticode:
@aldesantis I share your same feeling. I was a bit hesitant is adding `protect_from_connection_error` here as I think it's a problem that would be better to address directly in Solidus core, first because `protect_from_connection_error` is a private method, then because `#void_transaction!` was introduced only very recently, so if we did fine until that moment without the help of  `protect_from_connection_error`, so maybe it's not really a needed improvement?

aldesantis:
@spaghetticode yeah, I think that makes sense. Let's move forward without it and at some point we should probably extract it directly in the core.

aldesantis (approved):

PR #78, opened 2020-07-27
Refactor spec with fill_in_card helper

brchristian:
Following the example of [solidus_address_helper.rb](https://github.com/solidusio/solidus_stripe/blob/master/spec/support/solidus_address_helper.rb), this creates a card helper which makes for more concise specs. It can be shared across both V2 and V3 Elements implementations.

It also helps to clarify which card input data is simply the defaults, and which is data specific to that particular spec.

### DIFF
--- a/lib/solidus_stripe/testing_support/card_input_helper.rb
+++ b/lib/solidus_stripe/testing_support/card_input_helper.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-module SolidusCardHelper
+module SolidusCardInputHelper
   def fill_in_card(card = {})
     card[:number] ||= "4242 4242 4242 4242"
     card[:code] ||= "123"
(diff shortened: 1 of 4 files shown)


spaghetticode (approved):
@brchristian thanks for the cleanup! ✂️ 🧹

aldesantis:
@brchristian awesome, thanks! Do you think it would be possible to move this to `lib` and turn it into a helper that can also be used by the main app? I've just written a helper like yours in one of our apps and it would be good to have it available out of the box.

brchristian:
@aldesantis Done!

spaghetticode (approved):

aldesantis:
@brchristian thanks! One last small thing: can you move the helper to `lib/solidus_stripe/testing_support`? That's the convention we go with when creating factories/test helpers. Other than that, GTM!

brchristian:
@aldesantis No problem, updated the path!

Issue #74, opened 2020-07-14
Roadmap to remove ActiveMerchant dependency?


brchristian:
It made complete sense that the old `solidus_gateway` extension was built on top of `active_merchant`, because both of them are meant to support a large number of gateways with as uniform a style as possible.

However, more recently Solidus has deprecated `solidus_gateway` in favor of individual extensions that support one gateway each, and hence in the case of Stripe we have the new `solidus_stripe` extension.

However, because it was extracted from the original `solidus_gateway` code, `solidus_stripe` is still built on `active_merchant`, which has a number of downsides, including:

- Increasing the number dependencies
- Increasing the attack surface
- More code for developers to learn
- Greater complexity in debugging
- Inability to take advantage of more Stripe-specific functionality that other gateways lack

I wonder, would there be interest in this community in a PR that would remove the dependency on `active_merchant` in favor of directly utilizing the official `stripe` gem?

I'd be happy to help if this would be of interest. :)

spaghetticode:
@brchristian I think it makes sense to switch to the official Stripe gem, mostly because it has a much richer API than ActiveMerchant. For example, my feeling is that we currently can't succeed in https://github.com/solidusio/solidus_stripe/issues/26 with ActiveMerchant, we need to use Stripe API from their gem.

The transition will probably take some time, as we're currently supporting functionalities (deprecated V2 behavior, V3 Elements, Payment Intents, payment request button) that need to be transposed whenever possible, keeping the impact of the change on existing stores as small as possible. 

If someone else has different opinions please chime in, but I personally think that if you're interested in starting to work on the transition, you're more than welcome in doing so. Of course, let's get in touch if you need some more information, help, or coordination.

brchristian:
@spaghetticode Okay, that sounds good! I will look into it. I was inspired partly by our interest in #26.

To elaborate a little bit on my own motivations, our company is looking to implement a fairly sophisticated subscription system, including automatic renewals. The Stripe API supports autorenew, but for that to work we need stable customer accounts on the Stripe side, etc. (Hence my interest in #26.) My feeling is that basing the `solidus_stripe` extension on the Stripe gem would make it easier to extend the functionality in this way.

As for the transition, given that the V2 behavior is deprecated, I might look at supporting only V3 behavior to start with, and see where that takes me.

And yes, would love your help and coordination. I'll be happy to use this thread (or some other channel, if you prefer) to stay in touch once I begin to look into things.

[...]

Issue #26, opened 2020-01-30
Reuse Stripe profiles when customer adds a new credit card


spaghetticode:
When a `Spree::User` adds a new credit card and that user already has a saved credit card stored we should reuse their existing Stripe customer profile instead of creating a new one. 

This is a screenshot from a Stripe test account that shows a single email address with 3 different customer accounts, each with one single credit card:

<img width="1404" alt="Schermata 2020-01-30 alle 17 54 59" src="https://user-images.githubusercontent.com/141220/73471454-b08f8e00-4389-11ea-9bac-3755ba7b8e21.png">

This can be surely done when using the Stripe official gem, but we need to investigate if ActiveMerchant allows this.

brchristian:
We are looking at building out a subscription implementation on top of `solidus_stripe`, and persistent customer accounts are a critical first step, so adding a +1 to this. LMK if I can help with a PR...

brchristian:
It seems to me that in an ideal world, this extension would include a migration to add a `stripe_gateway_customer_id` to the `Spree::Users` table.

But, maybe there's another way to do it. When creating a new `Spree::CreditCard` object, we can attempt to populate the `gateway_customer_id` field like so:

```ruby
gateway_customer_profile_id = source.user&.
payment_sources&.where&.not(gateway_customer_profile_id: nil)&.order(:updated_at)&.last&.
gateway_customer_profile_id
```

@spaghetticode What do you think about trying something like that?

brchristian:
Updating with two pieces of information. First is that I've created a (failing) spec at #77 to demonstrate this issue.

Second is that I have begun tracking down the cause. I believe the key lines are here:

https://github.com/solidusio/solidus_stripe/blob/eca6dc226241116f75266a4d5c1bc0b789c62727/app/models/spree/payment_method/stripe_credit_card.rb#L122-L139

Here you can see how we're setting the `gateway_customer_profile_id`. Of course, a trivial way to make the specs in #77 pass would be to change the lines that directly update the payment source, but this would not be sufficient because the data would look right to Solidus and in the database, but would not actually be correct in Stripe.

It seems that we need to link this to the customer either before or during the call to [`gateway.store`](https://github.com/activemerchant/active_merchant/blob/b3d331256ca6d70374fc4cbca5d32dfd6857560e/lib/active_merchant/billing/gateways/stripe.rb#L193-L231). This method lives in ActiveMerchant. So, I'll keep digging and update this thread when I learn more.

[...]

PR #81, opened 2020-08-15
Refactor create_profile to use Stripe gem

brchristian:
This PR is a replacement / extension for the work in #77. It can commit cleanly if #77  is merged, or alternately #77 can be closed and this PR can take its place.

Changes from #77:

- Now [includes (passing) tests for maintaining stable Stripe customer profiles across purchases when using V3 Intents](https://github.com/solidusio/solidus_stripe/blob/a7d01ca16eb2fb918ac64d1e4b3afec7660c509e/spec/features/stripe_customer_spec.rb#L176-L181)! 🎉  In my view, that makes this a complete solution to #26.
- Uses the official Stripe gem rather than ActiveMerchant for the `create_profile` method, a stepping stone to the long-term goal of #74.

I’m very pleased with this result.

@spaghetticode Your code review of #77 was amazing, and I incorporated all of your suggestions there into this PR as well. What do you think?

### DIFF
--- a/app/decorators/models/spree/credit_card_decorator.rb
+++ b/app/decorators/models/spree/credit_card_decorator.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+module Spree
+  module CreditCardDecorator
+    def cc_type=(type)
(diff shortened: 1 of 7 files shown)


elia:

PR #82, opened 2020-08-20
Sync Solidus Users and Stripe Customers

brchristian:
Solves #26.
Begins the solution of #74.

NOTE: This PR is an alternative (IMO a cleaner and superior one) to both #77 and #81.

The idea is that the `gateway_customer_id` really represents a mapping between `Spree::Users` and `Stripe::Customers`. As [Stripe puts it](https://stripe.com/docs/payments/save-during-payment#web-create-a-customer): "You should create a Customer object when your customer creates an account with your business." This achieves that goal, with `Spree::Users` and `Stripe::Customer` now linked.

This allows us to create a very clean implementation of preserving the Stripe::Customer across multiple Solidus purchases (with both the same and different cards), across V2, Elements, and Intents. I believe it is an even better and more principled/elegant solution to #26 than either #77 and #81.

This also begins the process of shifting the `solidus_stripe` extension from its dependency on ActiveMerchant onto the official Stripe gem, which when complete will be I think a terrific simplification as well as giving us greater power and flexibility in how Solidus stores can take the fullest advantage of Stripe.


db/migrate/20200726022620_add_stripe_customer_id_to_users.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class AddStripeCustomerIdToUsers < SolidusSupport::Migration[5.1]
+  def up
+    add_column :spree_users, :stripe_customer_id, :string, unique: true
```
igorbp: There's a concern regarding this migration. On [custom authentications](https://guides.solidus.io/developers/users/custom-authentication.html), the table may not be `:spree_users` (or Spree::User model). This migration would need to find the correct user class through `Spree.user_class`.

db/migrate/20200726022620_add_stripe_customer_id_to_users.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    Spree::User.includes(bill_address: :country, ship_address: :country).find_each do |u|
+      user_stripe_payment_sources = user&.wallet&.wallet_payment_sources&.select do |wps|
```
igorbp: I think `user` here is meant to be `u` or probably the other way around, isn't it?

igorbp (commented):
Hi @brchristian. This looks like a good start toward supporting the official stripe gem :)

I've added a few comments regarding the migration.

db/migrate/20200726022620_add_stripe_customer_id_to_users.rb
brchristian: Good point! I changed `:spree_users` to `Spree.user_class.table_name.to_sym`.

db/migrate/20200726022620_add_stripe_customer_id_to_users.rb
brchristian: Good catch, thank you! I changed it to `u`.

brchristian:
Thanks @igorbp! I appreciate the careful look. I've made both of those suggested changes and have force-pushed this PR.

config/initializers/solidus_stripe.rb
```diff
@@ -0,0 +1 @@
+Stripe.api_key = Spree::PaymentMethod::StripeCreditCard.last&.preferences&.dig(:secret_key)
```
igorbp, line 1: The Stripe `secret_key` can also be defined in the database, which means it could be changed in the runtime. In that case, this would still have the old key. It would be better to define this during the runtime. This is also a good indication that we might need a class to handle the integration.

db/migrate/20200726022620_add_stripe_customer_id_to_users.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class AddStripeCustomerIdToUsers < SolidusSupport::Migration[5.1]
+  def up
+    add_column Spree.user_class.table_name.to_sym, :stripe_customer_id, :string, unique: true
+
+    Spree::User.includes(bill_address: :country, ship_address: :country).find_each do |u|
```
igorbp, line 7: This part could be in a different migration. I believe with the custom user class situation, some users would want to deal with it themselves, and would be easier to skip a migration instead of comment this one.

igorbp (commented):
Hey @brchristian. Thanks for considering my comments last time. :) 

I made a few more suggestions that you might want to take a look at. I would be glad to hear your opinion about it.

elia:

PR #85, opened 2020-10-01
Add missing 'var'

willread:
The `onShippingAddressChange` function is missing a `var` keyword which results in an implicit global and causes a strict mode violation.

There doesn't appear to be any code depending on this behaviour so I assume this is a simple fix.


spaghetticode (approved):
@willread thanks for the fix 👍

spaghetticode:
CI failure is unrelated to the changes, so I'm merging anyway.

PR #86, opened 2020-10-02
Fix 3DS iframe selection

spaghetticode:
This fix should bring the gem CI back to green status.

Stripe is now using random names for their iframe containers, so we need to resort to some other strategy for finding the iframe that contains the one named `__stripeJSChallengeFrame`.

Currently, one reliable way to find this iframe is by using a non-changing part of its `src` attribute.

### DIFF
--- a/spec/features/stripe_checkout_spec.rb
+++ b/spec/features/stripe_checkout_spec.rb
@@ -471,7 +471,7 @@ RSpec.describe "Stripe checkout", type: :feature do
   end
 
   def within_3d_secure_modal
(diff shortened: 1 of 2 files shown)


aldesantis:
@spaghetticode looks like the CI is still failing, but I'm not sure if the problem is related.

spaghetticode:
@aldesantis yes it is, that's because of some recent changes in Solidus... which I authored myself 🤪 
Hopefully, it's going to be green now.

aldesantis (approved):

PR #92, opened 2020-11-26
Remove unneeded braces in specs

brchristian:


chrean:
Thanks for the contribution, and sorry for the late response. We are working on overhauling the extension, so I am closing this because it will soon become irrelevant.

PR #95, opened 2021-01-15
Fix card input helper

filippoliverani:
This PR improves card input helper used in feature spec to make it work with current Stripe Elements html fields.


filippoliverani:
Moved into https://github.com/solidusio/solidus_stripe/pull/97

filippoliverani:
Moved into https://github.com/solidusio/solidus_stripe/pull/97

PR #99, opened 2021-01-22
Remove Solidus 2.x deprecation to allow 3.0 usage

kennyadsl:
Closes #98 by removing all the deprecation warnings making this gem compatible with Solidus 3.0.

Blocked by https://github.com/solidusio/solidus_stripe/pull/97 and https://github.com/solidusio/solidus/pull/3908.

Co-authored-by: @rainerdema

PR #100, opened 2021-01-29
Maintain compatibility with Solidus 3.0 Address#name

brchristian:
Solidus 3.0 switched from `Address#first_name` and `Address#last_name` to simply `Address#name` -- and also removed the `Address#full_name` accessor for good measure. (See https://github.com/solidusio/solidus/pull/3458)

This caused the `solidus_stripe` specs to fail on the Solidus `master` branch.

This PR accommodates the new Address format in Solidus 3.0 while maintaining backwards compatibility with 2.11, 2.10, and older versions of Solidus. It fixes the specs which are currently broken in Solidus 3.0.

### DIFF
--- a/app/models/solidus_stripe/address_from_params_service.rb
+++ b/app/models/solidus_stripe/address_from_params_service.rb
@@ -12,7 +12,13 @@ module SolidusStripe
       if user
         user.addresses.find_or_initialize_by(attributes)
       else
-        Spree::Address.new(attributes)
+        begin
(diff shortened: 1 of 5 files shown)


brchristian:
Note: This may already have been fixed by #99. If so, I fully support the merging of #99, which has an alternate implementation to this PR and which also addresses a number of deprecation warnings.

kennyadsl:
Thanks anyway @brchristian. I appreciate the help effort!

PR #105, opened 2021-04-22
Update install instructions

kennyadsl:
I can't see any value in specifying the version in the README, also, we can use `bundle add` that simplifies the instructions.


spaghetticode (approved):
@kennyadsl I agree, thanks!

Issue #64, opened 2020-05-31
"You cannot cancel this PaymentIntent because it has a status of canceled" error


stuffmatic:
I've started seeing an error that I haven't seen before when trying to proceed from the payment step:

_Spree::Core::GatewayError (You cannot cancel this PaymentIntent because it has a 
status of canceled. Only a PaymentIntent with one of the following statuses may be 
canceled: requires_payment_method, requires_capture, requires_confirmation, requires_action.):_

As far as I can tell, I haven't made any changes to my code that would explain this and I'm using a test card number that I've successfully used before (4242 4242 4242 4242).

Here's the [full stacktrace](https://github.com/solidusio/solidus_stripe/files/4707465/stacktrace.txt)

I'm using solidus_stripe 4.0.0, solidus 2.10 and rails 6.0.2.2.

Any clues would be greatly appreciated.

stuffmatic:
Update: I tried the checkout flow in a private window and it seems to work, so maybe I've got some stale cookie data or something. Regardless, it would be nice if the error above is propagated to the frontend so it can be displayed. Currently nothing happens after the crash and the frontend remains "locked" until manually reloaded.

spaghetticode:
Hi @stuffmatic, thanks for sharing this issue. I have never seen that error before payment complete, and from what you wrote it seems to be related to your specific session, so it may be hard to reproduce.

FYI I fixed the very same error recently when refunding an order, see https://github.com/solidusio/solidus_stripe/issues/56.  The fix is still only in master, so you may want to bundle that branch until a new release is created.

stuffmatic:
@spaghetticode Yeah, this seems to be an unlikely corner case, so I'm not too concerned about this happening. However, it would be nice if the error was passed to the frontend in the same way as other errors (payment declined etc). Currently, it seems like `SolidusStripe::IntentsController#create_payment` is crashing with a status 500.

spaghetticode:
@stuffmatic what you say about the 500 error makes sense, we need to avoid that the FE interface stops being responsive under those circumstances. A simple solution that comes to my mind is we can rescue `Spree::Core::GatewayError` in `CreateIntentsPaymentService#call` and make it return false. Not really sure if we want to expose the original Stripe error message to the FE as well.

By the way, did you happen to see more instances of that error?

stuffmatic:
> By the way, did you happen to see more instances of that error?

[...]

PR #115, opened 2021-09-12
Fix payment_intent_unexpected_state error when voiding payment intents

jfturcot:
Solves #64

If a payment intent has been created but the order is never completed, Stripe will eventually time out the intent and change its state. When a new payment intent is created, it first tries to void all the previous uncompleted payment intents of the user and this would break with the Stripe error payment_intent_unexpected_state since Stripe's state machine does not allow that change.

### DIFF
--- a/app/decorators/models/spree/payment_decorator.rb
+++ b/app/decorators/models/spree/payment_decorator.rb
@@ -6,6 +6,18 @@ module Spree
       gateway_order_id
     end
 
+    def handle_void_response(response)
+      record_response(response)
+
+      if response.success? ||
+         (response.params['error'] && response.params['error']['code'] == 'payment_intent_unexpected_state')
+        self.response_code = response.authorization
+        void
+      else
+        gateway_error(response)
+      end
+    end
+
     ::Spree::Payment.prepend(self)
   end
 end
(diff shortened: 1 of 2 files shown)


kennyadsl (commented):
Thanks for your contribution @jfturcot!

I was wondering if there's a way to add this logic to [`SolidusStripe::PaymentMethod` ](https://github.com/solidusio/solidus_stripe/blob/b7dfc00173be4dc35e605c79e77448a45a2fc666/app/models/solidus_stripe/payment_method.rb)instead, in order to avoid overriding the payment class (which will be used by other payment methods as well).

The class I referenced has `Spree::PaymentMethod` as parent class, which implements a generic [`try_void`](https://github.com/solidusio/solidus/blob/248448132352abc158f4bcff730cb8aea1599173/core/app/models/spree/payment_method.rb#L169-L192) method, which is called when the subclass doesn't implement one. 

What do you think? Would it make sense to move this logic there?

kennyadsl:
Hello @jfturcot! Are you still interested in moving this PR forward? Otherwise, we can probably continue the work to fix the issue. Thanks!

jfturcot:
@kennyadsl Sure! Sorry I have been very busy lately, but I would love to get this merged. I will take a look at it over the weekend.

fthobe:
@jfturcot any chance you pick this up again?

elia:

PR #121, opened 2022-02-24
allow stripe connect options when create intent

ccarruitero:
This PR is sponsored by [MagmaLabs](https://www.magmalabs.io/)

This allow `solidus_stripe` gem to use [stripe connect](https://stripe.com/docs/connect).

Stripe connect allows you to split payment, so you can take a fee from there. The main use for these feature is for marketplaces or platforms.

This implement the stripe connect with [direct](https://stripe.com/docs/connect/direct-charges) or [destination](https://stripe.com/docs/connect/destination-charges) charges.

I'm not totally sure where will be the best place to set the `connected_account` in the gem. This will be the account where stripe transfer the payment less the fee. 
For now I set that in payment method, but maybe will be better left that part to the application implementation.

Let me know what do you think?

### DIFF
--- a/app/controllers/solidus_stripe/intents_controller.rb
+++ b/app/controllers/solidus_stripe/intents_controller.rb
@@ -50,16 +50,14 @@ module SolidusStripe
     end
 
     def create_payment_intent
+      intent_options = SolidusStripe::PrepareOptionsForIntentService.call(
+        current_order,
+        stripe
+      )
       stripe.create_intent(
         (current_order.total * 100).to_i,
         params[:stripe_payment_method_id],
-        description: "Solidus Order ID: #{current_order.number} (pending)",
-        currency: current_order.currency,
-        confirmation_method: 'automatic',
-        capture_method: 'manual',
-        confirm: true,
-        setup_future_usage: 'off_session',
-        metadata: { order_id: current_order.id }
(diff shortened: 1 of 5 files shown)


ericgross:
@elia @ccarruitero I am interested in using this (or a similar) enhancement to collect connect payments. I'm curious if you are using this in production and if you might have made any changes since pushing this up.

Is this something that could be merged or is it waiting on anything? I can try to help if there is anything that's needed.

Thanks again for all this work.

ccarruitero:
Hi @ericgross 

I was looking into [stripe connect docs](https://docs.stripe.com/connect) and I think the intent options for connect is still the same

But `solidus_stripe` gem changed a bit since this PR was opened. Right now this PR is not compatible with latest version.

Looking into solidus_stripe, it already [handle additional parameters when create intent](https://github.com/solidusio/solidus_stripe/blob/main/app/models/solidus_stripe/payment_intent.rb#L18). So, if you want to reimplement this, you only need to add the logic for connect options in the [gateway authorize method](https://github.com/solidusio/solidus_stripe/blob/ede6ec937b79ee6098cddfed25d9c6b7da7ea92e/app/models/solidus_stripe/gateway.rb#L42)

elia:

Issue #127, opened 2022-09-11
ActionView::MissingTemplate in Spree::Checkout#edit


joyboy67:
**Solidus version**

3.2.2 with solidus_starter_frontend

**Problem**

Hello I installed solidus_stripe (4.3.0) but when I go to the route `checkout/payment` I get

```
ActionView::MissingTemplate in Spree::Checkout#edit 

 Showing /home/computer/shop/app/views/spree/checkout/steps/payment_step/_payment_method_controls.html.erb where line#4 raised:

  Missing partial spree/checkout/payment/_stripe with {:locale=>[:fr], :formats=>[:html], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby, :jbuilder, :haml]}.

Searched in:
  * "/home/computer/shop/app/views"
  * "/usr/local/lib/ruby/gems/3.1/gems/solidus_stripe-4.3.0/lib/views/backend"
  * "/usr/local/lib/ruby/gems/3.1/gems/solidus_stripe-4.3.0/lib/views/api"
  * "/usr/local/lib/ruby/gems/3.1/gems/view_component-2.71.0/app/views"
  * "/usr/local/lib/ruby/gems/3.1/gems/solidus_backend-3.2.2/app/views"
  * "/usr/local/lib/ruby/gems/3.1/gems/kaminari-core-1.2.2/app/views"
  * "/usr/local/lib/ruby/gems/3.1/gems/solidus_api-3.2.2/app/views"
  * "/usr/local/lib/ruby/gems/3.1/gems/solidus_auth_devise-2.5.4/lib/views/backend"
  * "/usr/local/lib/ruby/gems/3.1/gems/devise-4.8.1/app/views"
  * "/usr/local/lib/ruby/gems/3.1/gems/solidus_core-3.2.2/app/views"
  * "/usr/local/lib/ruby/gems/3.1/gems/actiontext-7.0.4/app/views"
  * "/usr/local/lib/ruby/gems/3.1/gems/actionmailbox-7.0.4/app/views"

Did you mean?  spree/api/payments/source_views/stripe
               spree/admin/log_entries/stripe
               spree/admin/payments/source_forms/stripe
               spree/admin/payments/source_views/stripe
               spree/checkout/payment/gateway
               spree/admin/promotions/rules/store
    
  
Extracted source (around line #4):

<% methods.each do |method| %>
     <li id="payment_method_<%= method.id %>">
       <%= render( 
[...]

PR #129, opened 2022-09-13
Revert "Add back custom view paths that were mistakenly removed"

elia:
Reverts solidusio/solidus_stripe#128

Turns out this is already taken care by solidus_support and https://github.com/solidusio/solidus_stripe/issues/127 is due to something else. I'm opening this revert PR as a reminder that we need to address solidus_starter_frontend support.

### DIFF
--- a/lib/solidus_stripe/engine.rb
+++ b/lib/solidus_stripe/engine.rb
@@ -16,10 +16,6 @@ module SolidusStripe
       g.test_framework :rspec
     end
 
-    paths["app/views"] << "lib/views/backend" if SolidusSupport.backend_available?
-    paths["app/views"] << "lib/views/frontend" if SolidusSupport.frontend_available?
-    paths["app/views"] << "lib/views/api" if SolidusSupport.api_available?
-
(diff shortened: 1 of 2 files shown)


kennyadsl:
There's some parts of this PR worth keeping?

joyboy67:
https://github.com/solidusio/solidus_stripe/issues/127#issuecomment-1249063818

elia:
@kennyadsl I think we can have a patchy solution adding a generator that copies all the frontend views and controller to the application's working dir which can be used as a temporarily fallback. The generator would be exactly what @joyboy67 did manually, this wouldn't solve any incompatibilities with SSF or any custom frontend an app could have but it's better than nothing. Until we figure out a more comprehensive solution.

@kennyadsl @joyboy67 thoughts?

_PS. I updated the commit to only remove the forced paths, keeping the fixes to the rakefile and CI config_

kennyadsl:
@elia nice, but I'll treat those things as separate concerns. I think we still need to revert this and after that, we should find a more general solution to this issue, which we have in all extensions (and I think it is exactly what we have been working on lately). Personally, I'd skip the patchy solution if possible.

kennyadsl (approved):

PR #139, opened 2022-12-05
Fix incorrect charge amount for currencies without fractions

### LINKED ISSUE #138: Fix incorrect charge amount for currencies without fractions
There is an issue with charged amounts in currencies without fractions (e.g. JPY) being divided by 100, which happens here: https://github.com/activemerchant/active_merchant/blob/db0e8d4f042b58aaff4af9fefef232f8e7528b8d/lib/active_merchant/billing/gateway.rb#L279
Perhaps override this method: https://github.com/solidusio/solidus/blob/e878076f2ed670d07654ab6293a16588743f2fa6/core/app/models/spree/payment.rb#L74

Error on Stripe API with fake card in testing mode:
image

Summary
I’m using the V1 api for a store in Japanese (JPY currency). Amounts are losing two zeros before being sent to the Stripe/Charges API.

Concerned Gems:
```
gem 'solidus', '~> 3.1'
gem 'solidus_stripe', 4.3'
```

Config:
```
  config.static_model_preferences.add(
    Spree::PaymentMethod::StripeCreditCard,
<img width="1478" alt="image" src="https://user-images.githubusercontent.com/7152249/205527744-9e2638e4-1ea4-4034-abbd-579032fa194f.png">

    'stripe_env_credentials',
    secret_key: ENV['STRIPE_SECRET_KEY'],
    publishable_key: ENV['STRIPE_PUBLISHABLE_KEY'],
    stripe_country: 'JP',
    v3_elements: false,
    v3_intents: false
  )
```

  My Research:
The auth trigger happens [here](https://github.com/solidusio/solidus/blob/e878076f2ed670d07654ab6293a16588743f2fa6/core/app/models/spree/payment/processing.rb#L37) (called by the state machine)
Sent to the gateway [here](https://github.com/solidusio/solidus/blob/e878076f2ed670d07654ab6293a16588743f2fa6/core/app/models/spree/payment/processing.rb#L37)
Money is calculated [here](https://github.com/solidusio/solidus/blob/e878076f2ed670d07654ab6293a16588743f2fa6/core/app/models/spree/payment.rb#L82)
It goes through Monetize / RubyMoney using the 'cents' method. That flows through to [here](https://github.com/RubyMoney/money/blob/ca59ced949e4e818262a59ffbcad3ec54affa81f/lib/money/money.rb#L658)

This can be seen in practice with something like this:
order = Spree::Order.find('ORDER NUMBER HERE')
amount = order.payments.first.amount
currency = order.currency
Spree::Money.new(amount, {currency: currency})
=> #<Spree::Money:0x000000013b0d9f58 @money=#<Money fractional:14000 currency:JPY>, @options={:sign_before_symbol=>true, :currency=>"JPY"}>
Spree::Money.new(amount, {currency: currency}).money.cents => 14000
 All looks good so far…
This is then processed by the gateway [here](https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/gateways/stripe.rb#L83)
Turned into a post for the Stripe API starting [here](https://github.com/activemerchant/active_merchant/blob/7fea8f3364ef44187aafa9c5eac3b76438671327/lib/active_merchant/billing/gateways/stripe.rb#L363)
The amount for the post is calculated [here](https://github.com/activemerchant/active_merchant/blob/7fea8f3364ef44187aafa9c5eac3b76438671327/lib/active_merchant/billing/gateways/stripe.rb#L403)
[The amount is localized here](https://github.com/activemerchant/active_merchant/blob/db0e8d4f042b58aaff4af9fefef232f8e7528b8d/lib/active_merchant/billing/gateway.rb#L279)

We can see that JPY is both in the payment (if you interrupt the authorization process to debug, the currency and country and correctly applied), so ‘JPY’ triggers currencies without fractions, which are specified [here](https://github.com/activemerchant/active_merchant/blob/db0e8d4f042b58aaff4af9fefef232f8e7528b8d/lib/active_merchant/billing/gateway.rb#L129%E2%80%A8).

This all posts to [the /charges/ path of the Stripe API](https://stripe.com/docs/api/charges/create)

  So, I think the amount is being converted to a float, then divided by 100, and then plugged into the api post. For example I have a product that has tax and shipping included that comes to ¥3200. At checkout in dev using a test api key, when going from confirm to complete (as authorize gets called), it’s giving an error that the amount is too low to be charged—¥32.
  
[I first brought this up on Slack, and was encouraged to give it a go myself](https://solidusio.slack.com/archives/C0JBKDF35/p1669932922971529).
I'm still working on trying to fix it myself, but it's proving to be [pretty stubborn](https://github.com/solidusio/solidus_stripe/pull/137) (almost the same text as here).

cmbaldwin:
## Summary

Fixes #137 
Fixes #138 

This feels really hacky—I'm sure there's a more elegant solution, but this works in my testing environment. I can't figure out why sometimes the 'money' argument is an integer and sometimes is a Spree:Money object in my testing. Maybe it was something else I was messing with (trying to override the 'money' method in Spree::Payment at one point). 

This also provides a Spree::Money object to ActiveMerchant::Billing::Gateway, but it looks like it doesn't want that anymore?https://github.com/activemerchant/active_merchant/blob/7fea8f3364ef44187aafa9c5eac3b76438671327/lib/active_merchant/billing/gateway.rb#L256

### DIFF
--- a/app/models/spree/payment_method/stripe_credit_card.rb
+++ b/app/models/spree/payment_method/stripe_credit_card.rb
@@ -61,20 +61,29 @@ module Spree
         true
       end
 
+      def non_fractional_money(money, currency)
+        amount = money.is_a?(Integer) ? money : money.cents
+        Spree::Money.new((amount * 100), { currency: currency })
+      end
+
+      def localize(money, currency)
+        ActiveMerchant::Billing::Gateway.currencies_without_fractions.include?(currency) ? non_fractional_money(money, currency) : money
+      end
+
       def purchase(money, creditcard, transaction_options)
-        gateway.purchase(*options_for_purchase_or_auth(money, creditcard, transaction_options))
+        gateway.purchase(*options_for_purchase_or_auth(localize(money, transaction_options[:currency]), creditcard, transaction_options))
       end
 
       def authorize(money, creditcard, transaction_options)
-        gateway.authorize(*options_for_purchase_or_auth(money, creditcard, transaction_options))
+        gateway.authorize(*options_for_purchase_or_auth(localize(money, transaction_options[:currency]), creditcard, transaction_options))
       end
 
       def capture(money, response_code, transaction_options)
-        gateway.capture(money, response_code, transaction_options)
+        gateway.capture(localize(money, transaction_options[:currency]), response_code, transaction_options)
       end
 
-      def credit(money, _creditcard, response_code, _transaction_options)
-        gateway.refund(money, response_code, {})
+      def credit(money, _creditcard, response_code, transaction_options)
+        gateway.refund(localize(money, transaction_options[:currency]), response_code, {})
       end
 
       def void(response_code, _creditcard, _transaction_options)
--- a/spec/models/spree/payment_method/stripe_credit_card_spec.rb
+++ b/spec/models/spree/payment_method/stripe_credit_card_spec.rb
@@ -171,9 +171,29 @@ describe Spree::PaymentMethod::StripeCreditCard do
     end
   end
 
+  context 'purchasing with fractionless currency' do
+    after do
+      subject.purchase(1999, 'credit card', { currency: 'JPY' })
+    end
+
+    it 'send the payment to the gateway' do
+      expect(gateway).to receive(:purchase).with('money', 'cc', 'opts')
+    end
+  end
+
   context 'authorizing' do
     after do
       subject.authorize(19.99, 'credit card', {})
+    end
+
+    it 'send the authorization to the gateway' do
+      expect(gateway).to receive(:authorize).with('money', 'cc', 'opts')
+    end
+  end
+
+  context 'authorizing with fractionless currency' do
+    after do
+      subject.authorize(1999, 'credit card', { currency: 'JPY' })
     end
 
     it 'send the authorization to the gateway' do
@@ -188,6 +208,20 @@ describe Spree::PaymentMethod::StripeCreditCard do
 
     it 'convert the amount to cents' do
       expect(gateway).to receive(:capture).with(1234, anything, anything)
+    end
+
+    it 'use the response code as the authorization' do
+      expect(gateway).to receive(:capture).with(anything, 'response_code', anything)
+    end
+  end
+
+  context 'capturing with fractionless currency' do
+    after do
+      subject.capture(1234, 'response_code', { currency: 'JPY' })
+    end
+
+    it 'amount not converted to cents' do
+      expect(gateway).to receive(:capture).with(Spree::Money.new(123_400, { currency: 'JPY' }), anything, anything)
     end
 
     it 'use the response code as the authorization' do


app/models/spree/payment_method/stripe_credit_card.rb
```diff
@@ -61,16 +61,26 @@ def payment_profiles_supported?
         true
       end
 
+      def non_fractional_money(money, currency)
+        amount = money if money.is_a?(Integer)
+        amount ||= money.cents if money.respond_to?(:cents)
+        Spree::Money.new((amount * 100), { currency: currency })
```
Naokimi: what would happen if money is neither an Integer nor it responds to :cents?
if that case can never happen, how about using a ternary operator? :)

cmbaldwin: Good point! [It shouldn't ever happen](https://github.com/solidusio/solidus/blob/e878076f2ed670d07654ab6293a16588743f2fa6/core/app/models/spree/payment.rb#L82). Thanks. Could be made into a one liner, but two lines seems more clear to me.

Also, in a rush I removed this fix from the 'credit' method after CI failed (I forgot to delete the underscore in the formerly unused param), but it needs to be fixed too because the 'refund' method (calculated [here](https://github.com/solidusio/solidus/blob/e878076f2ed670d07654ab6293a16588743f2fa6/core/app/models/spree/payment.rb#L103) in Solidus) still goes through [here](https://github.com/activemerchant/active_merchant/blob/7fea8f3364ef44187aafa9c5eac3b76438671327/lib/active_merchant/billing/gateway.rb#L279) in ActiveMerchant.

elia (commented):
@cmbaldwin Love the change! Any chance you could add specs and squash fixup commits to their original?

Happy to help in case it's not to clear how to do either things 🙌

cmbaldwin:
Also two new projects for me, but I gave it a shot. The squash seems fine, but I'm not very confident with the spec, is this OK?

elia:
Not sure why the status checks are not reported, but it seems to be green ✅  https://app.circleci.com/pipelines/github/solidusio/solidus_stripe/520/workflows/698cf6e8-5db4-4773-85b5-b165b7461e1d

elia (approved):

elia:
@cmbaldwin tried a force-push in order to trigger status checks

cmbaldwin:
Weird, I signed up for Circle CI to learn more about it. When I go to the branch to see the status it says:
`solidus_stripe 9 Orb solidusio/extensions@volatile not loaded. To use this orb, an organization admin must opt-in to using third party orbs in Organization Security settings.`
Maybe when I signed up I took ownership of the branch, and this somehow interrupts the normal process? 
I unfollowed it and clicked "stop building" on my end, to try and fix it.

elia:
CircleCI looks broken on this PR, I verified manually that it actually ran and succeeded, so I'm force-merging.

PR #156, opened 2022-12-30
MVP implementation that can complete a checkout

### LINKED ISSUE #152: Implement the standard gateway interfaces using the Stripe gem
There is a number of standard methods that should be implemented on the payment gateway as described in this guide: https://guides.solidus.io/advanced-solidus/payments-and-refunds/#custom-payment-gateways.

The gateway is expected to implement these methods using the Stripe gem (and not ActiveMerchant):

- [x] #initialize(options): initializes the gateway with the provided options. By default, Solidus will pass the [payment method's preferences](https://github.com/solidusio/solidus/blob/v3.0/core/app/models/spree/payment_method.rb#L78) in here.
- [x] #authorize(money, source, options = {}): authorizes a certain amount on the provided payment source.
- [x] #capture(money, transaction_id, options = {}): captures a certain amount from a previously authorized transaction.
- [x] #purchase(money, source, options = {}): authorizes and captures a certain amount on the provided payment source.
- [x] #void(transaction_id, [source,] options = {}): voids a previously authorized transaction, releasing the funds that are on hold. The source parameter is only needed for payment gateways that support payment profiles.
- [x] #credit(money, [source,] transaction_id, options = {}): refunds the provided amount on a previously captured transaction. The source parameter is only needed for payment gateways that support payment profiles.

elia:
## Summary

Aiming at the simplest implementation that allows completing a checkout.
The mapping between Stripe and Solidus models won't be final and subject to deep changes.
We don't expect anyone grabbing this extension from master given the big warning against that that we have at the top of the readme.

Fixes #152 

## Checklist

- [ ] I have used clear, explanatory commit messages.

The following are not always needed (~cross them out~ if they are not):

- [ ] I have added automated tests to cover my changes.
- [ ] I have attached screenshots to demo visual changes.
- [ ] I have opened a PR to update the [guides](https://github.com/solidusio/edgeguides).
- [ ] I have updated the README to account for my changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,16 @@ bin/rails generate solidus_stripe:install
 
 ## Development
 
+Retrieve your API Key and Publishable Key from your [Stripe testing dashboard](https://stripe.com/docs/testing).
+
+Set `SOLIDUS_STRIPE_API_KEY` and `SOLIDUS_STRIPE_PUBLISHABLE_KEY` environment variables (e.g. via `direnv`), this
+will trigger the default initializer to create a static preference for SolidusStripe.
+
+Run `bin/dev` to start both the sandbox rail server and the file watcher that will update the sandbox whenever
+a file is changed.
+
+Visit `/admin/payments` and create a new Stripe payment using the static preferences.
+
 ### Testing the extension
 
 First bundle your dependencies, then run `bin/rake`. `bin/rake` will default to building the dummy
--- a/app/models/solidus_stripe/gateway.rb
+++ b/app/models/solidus_stripe/gateway.rb
@@ -6,7 +6,6 @@ require 'stripe'
 # - https://stripe.com/docs/payments/accept-a-payment?platform=web&ui=checkout#auth-and-capture
 # - https://stripe.com/docs/charges/placing-a-hold
 module SolidusStripe
-  # rubocop:disable Lint/UnusedMethodArgument
   class Gateway
     def initialize(options)
       # Cannot use kwargs because of how the Gateway is initialized by Solidus.
@@ -19,6 +18,8 @@ module SolidusStripe
     attr_reader :client
 
     # Authorizes a certain amount on the provided payment source.
+    #
+    # @see #purchase
     def authorize(amount_in_cents, source, options = {})
       payment_intent_options = options[:payment_intent_options].dup.to_h
       payment_intent_options[:capture_method] = "manual"
@@ -28,7 +29,18 @@ module SolidusStripe
     # Captures a certain amount from a previously authorized transaction.
     # Ref: https://stripe.com/docs/api/payment_intents/capture#capture_payment_intent
     # Ref: https://stripe.com/docs/payments/capture-later
-    def capture(amount_in_cents, _transaction_id, options = {})
+    #
+    # @todo add support for capturing custom amounts
+    #
+    # @param _amount_in_cents [Integer] The fractional amount as defined by Spree::Money,
+    #   although it's cents for most currencies some will have a different multiplier.
+    #   Currently ignored (see "todo" section).
+    # @param _transaction_id [Object] (currently ignored)
+    # @param options [Hash]
+    # @option options [Spree::Payment] :originator the payment on which the #capture is being performed (required)
+    #
+    # @return ActiveMerchant::Billing::Response
+    def capture(_amount_in_cents, _transaction_id, options = {})
       payment = options[:originator] or raise ArgumentError, "please provide a payment with the :originator option"
       payment_intent_id = payment.source.stripe_payment_intent_id
 
@@ -50,7 +62,15 @@ module SolidusStripe
     end
 
     # Authorizes and captures a certain amount on the provided payment source.
-    def purchase(amount_in_cents, source, options = {})
+    #
+    # @param amount_in_cents [Integer] The fractional amount as defined by Spree::Money,
+    #   although it's cents for most currencies some will have a different multiplier.
+    # @param source [Spree::PaymentSource, nil] optionally, payment source from which to create the authorization
+    # @param options [Hash]
+    # @option options [Hash] :payment_intent_options options forwarded to `Stripe::PaymentIntent.create`
+    #
+    # @return ActiveMerchant::Billing::Response
+    def purchase(amount_in_cents, _source, options = {})
       currency = options.fetch(:currency)
 
       # Charge the Customer instead of the card:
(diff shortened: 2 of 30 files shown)


app/models/solidus_stripe/payment_source.rb
```diff
@@ -4,5 +4,22 @@
 
 module SolidusStripe
   class PaymentSource < ::Spree::PaymentSource
+    def gateway_payment_profile_id=(value)
```
kennyadsl: Thoughts on using https://apidock.com/rails/Module/alias_attribute?

app/models/solidus_stripe/gateway.rb
```diff
@@ -13,31 +16,90 @@ def initialize(options)
...
+      payment_intent_id = payment.source.stripe_payment_intent_id
+
+      raise ArgumentError, "missing transaction_id" unless payment_intent_id
+
+      unless payment_intent_id.start_with?('pi_')
+        raise ArgumentError, "the payment-intent id has the wrong format"
+      end
+
+      # To create a requires_capture PaymentIntent,
+      # see our guide at: https://stripe.com/docs/payments/capture-later
```
kennyadsl: Is `our` referring to Stripe here? Maybe we can be more explicit about it.

README.md
```diff
@@ -43,6 +43,16 @@ bin/rails generate solidus_stripe:install
 
 ## Development
 
+Retrieve your API Key and Publishable Key from your [Stripe testing dashboard](https://stripe.com/docs/testing).
+
+Set `SOLIDUS_STRIPE_API_KEY` and `SOLIDUS_STRIPE_PUBLISHABLE_KEY` environment variables (e.g. via `direnv`), this
+will trigger the default initializer to create a static preference for SolidusStripe.
+
+Run `bin/dev` to start both the sandbox rail server and the file watcher that will update the sandbox whenever
```
kennyadsl, line 51: What if `bin/dev` also creates the payment method if needed? I think it would be a great UX improvement but don't know if that's easy to do. Thoughts?

app/models/solidus_stripe/payment_method.rb
```diff
@@ -25,6 +25,20 @@ def gateway_class
       Gateway
     end
 
+    def create_profile(payment)
+      pi = payment.source.payment_intent
```
kennyadsl: I'd prefer to avoid 2-letter variables. `payment_intent` is not that long that will create problems IMHO, thoughts?

spec/requests/solidus_stripe/payment_intents_controller_spec.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+require 'solidus_stripe_spec_helper'
+
+RSpec.describe SolidusStripe::PaymentIntentsController, type: :request do
```
kennyadsl, line 5: If I'm not mistaken here, the best practice is to avoid adding the `_controller` part to the filename and to the main describe name. In general, the file seems too tied to the controller, while requests specs should be more of a full request-cycle spec.

lib/generators/solidus_stripe/install/templates/app/views/checkouts/payment/_stripe.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  data-solidus-stripe-payment-solidus-payment-method-id-value="<%= payment_method.id %>"
+  data-solidus-stripe-payment-payment-intent-path-value="<%= solidus_stripe.payment_intents_path %>"
+  data-solidus-stripe-payment-email-address-value="<%= current_order.email %>"
+  data-solidus-stripe-payment-payment-url-value="<%= checkout_state_url(:payment) %>"
+  data-solidus-stripe-payment-payment-form-selector-value="#checkout_form_payment"
+  data-action="submit@window->handleSubmit"
+>
+  <div data-solidus-stripe-payment-target="paymentElement">
+    <!-- Elements will create form elements here -->
+    <em>Loading...</em>
```
kennyadsl: Good candidate for `I18n`?

app/views/spree/admin/payments/source_views/_stripe.html.erb
```diff
@@ -0,0 +1,16 @@
+<fieldset data-hook="credit_card">
+  <legend align="center"><%= SolidusStripe::PaymentSource.model_name.human %></legend>
+
+  <div class="row">
+    <div class="col-4">
+      <dl>
+        <dt><%= SolidusStripe::PaymentSource.human_attribute_name(:stripe_payment_intent_id) %>:</dt>
+        <dd><%= link_to(
+          payment.source.stripe_payment_intent_id,
+          "https://dashboard.stripe.com/#{'test/' if payment.payment_method.preferred_test_mode}payments/#{payment.source.stripe_payment_intent_id}",
```
kennyadsl: Thought about moving this URL generation to the payment source model?

kennyadsl (commented):
Thank @elia, this is a great start. I left just some questions/comments. 

Also, now that we are fresh with some concept, I think this might be a good moment to add some YARD documentation to the main gateway/payment method methods that could help us in the future. Do you think this might have some value?

README.md
elia: Added the static pref that for some reason was left out of the PR and bin/sandbox will auto-create the payment method connecting to the static pref if the same env var is present. 👍

spec/requests/solidus_stripe/payment_intents_controller_spec.rb
elia: I just checked the RSpec generator as I wasn't sure about what changes I added, this is how it looks like:

```rb
require 'rails_helper'

RSpec.describe "<%= class_name.pluralize %>", <%= type_metatag(:request) %> do
  describe "GET /<%= name.underscore.pluralize %>" do
    it "works! (now write some real specs)" do
      get <%= index_helper %>_path
      expect(response).to have_http_status(200)
    end
  end
end
```

So the only thing I changed was removing the quotes from around the class name so I could use `described_class`, happy to change it if there's a different/better way.

kennyadsl: Yes, but I think that the class should not be a controller in the request specs. Looking around for examples it seems like it's a more high level concept that a specific class.

kennyadsl:
@elia I think you missed the YARD documentation part of my review comment, I can't find any reply to that and nothing changed in the code proposed. 👓

elia:
> @elia I think you missed the YARD documentation part of my review comment, I can't find any reply to that and nothing changed in the code proposed. 👓

Thanks for calling that out, I indeed missed the comment.

It's probably early for adding documentation but I did it anyway mostly to have an idea of what it looks like, many gateway methods are still incomplete and will change considerably, but the current situation should be covered accurately.

kennyadsl (approved):
Thanks Elia, I'm sure that even if it's going to change, this Yard documentation won't be regretted.

PR #157, opened 2023-01-20
Add support for Stripe webhooks

waiting-for-dev:
## Summary

We set up the backbone to support [Stripe webhook](https://stripe.com/docs/webhooks) events. There're two different use cases:

- Users want to listen to certain webhook events. They'll need to add their names in a `webhook_events` setting.
- This extension code makes internal use of some of them. That's something programmed, and when the time comes, we'll need to add the event names in the `SolidusStripe::Webhook::Event::CORE_EVENTS` constant.

A new `POST /solidus_stripe/webhooks` endpoint exists to support the integration. Users must also configure a `solidus_stripe.webhook_endpoint_secret` Rails credential to [validate the incoming events](https://stripe.com/docs/webhooks/signatures). There's also a `webhook_signature_tolerance` setting to change the default number of seconds [while a signature is valid](https://stripe.com/docs/webhooks/signatures#replay-attacks).

When an incoming webhook arrives, a matching `Spree::Bus` event prepended with `stripe.` is published. Therefore, all the usual techniques in [Solidus's pub/sub system](https://guides.solidus.io/customization/subscribing-to-events) are allowed.

I'd like to move the decoding of the event from a controller to a service, but as we don't have the infrastructure ready on Solidus is better to wait and not implement a custom solution here.

## Checklist

- [x] I have used clear, explanatory commit messages.

The following are not always needed (~cross them out~ if they are not):

- [x] I have added automated tests to cover my changes.
- ~[ ] I have attached screenshots to demo visual changes.~
- ~[ ] I have opened a PR to update the [guides](https://github.com/solidusio/edgeguides).~
- [x] I have updated the README to account for my changes.

### DIFF
--- a/app/controllers/solidus_stripe/webhooks_controller.rb
+++ b/app/controllers/solidus_stripe/webhooks_controller.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require "solidus_stripe/webhook/event"
+require "stripe"
+
+module SolidusStripe
+  class WebhooksController < Spree::BaseController
+    SIGNATURE_HEADER = "HTTP_STRIPE_SIGNATURE"
+
+    skip_before_action :verify_authenticity_token, only: :create
+
+    respond_to :json
+
+    def create
+      event = Webhook::Event.from_request(payload: request.body.read, signature_header: signature_header)
+      return head(:bad_request) unless event
+
+      Spree::Bus.publish(event) && head(:ok)
+    end
+
+    private
+
+    def signature_header
+      request.headers[SIGNATURE_HEADER]
+    end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -2,4 +2,5 @@
 
 SolidusStripe::Engine.routes.draw do
   resources :payment_intents, only: :create
+  resources :webhooks, only: :create, format: false
 end
--- a/lib/solidus_stripe/configuration.rb
+++ b/lib/solidus_stripe/configuration.rb
@@ -1,10 +1,24 @@
 # frozen_string_literal: true
 
+require "stripe/webhook"
+
 module SolidusStripe
   class Configuration
-    # Define here the settings for this extension, e.g.:
-    #
-    # attr_accessor :my_setting
+    # @!attribute [rw] webhook_events
+    #  @return [Array<Symbol>] stripe events to handle. You also need to
+    #  register them in the Stripe dashboard. For an event `:foo`, a matching
+    #  `:"stripe.foo"` event will be registered in {Spree::Bus}.
+    attr_accessor :webhook_events
+
+    # @!attribute [rw] webhook_signature_tolerance
+    #  @return [Integer] number of seconds while a webhook event is valid after
+    #  its creation. Defaults to {Stripe::Webhook::DEFAULT_TOLERANCE}.
+    attr_accessor :webhook_signature_tolerance
+
+    def initialize
+      @webhook_events = []
+      @webhook_signature_tolerance = Stripe::Webhook::DEFAULT_TOLERANCE
+    end
   end
 
   class << self
--- a/lib/solidus_stripe/engine.rb
+++ b/lib/solidus_stripe/engine.rb
@@ -14,6 +14,16 @@ module SolidusStripe
       app.config.spree.payment_methods << 'SolidusStripe::PaymentMethod'
     end
 
+    initializer "solidus_stripe.pub_sub", after: "spree.core.pub_sub" do |app|
+      require "solidus_stripe/webhook/event"
+      app.reloader.to_prepare do
+        SolidusStripe::Webhook::Event.register(
+          user_events: SolidusStripe.configuration.webhook_events,
+          bus: Spree::Bus
+        )
+      end
+    end
+
     # use rspec for tests
     config.generators do |g|
       g.test_framework :rspec
--- a/lib/solidus_stripe/webhook/event.rb
+++ b/lib/solidus_stripe/webhook/event.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+module SolidusStripe
+  module Webhook
+    # Omnes event wrapping a Stripe event.
+    #
+    # All unknown methods are delegated to the wrapped Stripe event.
+    #
+    # @see Stripe::Event
+    class Event
+      include Omnes::Event
+
+      PREFIX = "stripe."
+      private_constant :PREFIX
+
+      # TBD
+      CORE_EVENTS = Set[*%i[]].freeze
+      private_constant :CORE_EVENTS
+
+      # @api private
+      class << self
+        def from_request(payload:, signature_header:, secret: default_secret, tolerance: default_tolerance)
+          stripe_event = Stripe::Webhook.construct_event(payload, signature_header, secret, tolerance: tolerance)
+          new(stripe_event: stripe_event)
+        rescue Stripe::SignatureVerificationError, JSON::ParserError
+          nil
+        end
+
+        def register(user_events:, bus:, core_events: CORE_EVENTS)
+          (core_events + user_events).each do |event|
+            bus.register(:"stripe.#{event}")
+          end
+        end
+
+        private
+
+        def default_tolerance
+          SolidusStripe.configuration.webhook_signature_tolerance
+        end
+
+        def default_secret
+          Rails.application.credentials.solidus_stripe[:webhook_endpoint_secret]
+        end
+      end
+
+      # @api private
+      attr_reader :omnes_event_name
+
+      # @api private
+      def initialize(stripe_event:)
+        @stripe_event = stripe_event
+        @omnes_event_name = :"#{PREFIX}#{stripe_event.type}"
+      end
+
+      # Serializable representation of the event.
+      #
+      # Ready to be consumed by async Omnes adapters, like
+      # {Omnes::Subscriber::Adapter::ActiveJob} or
+      # {Omnes::Subscriber::Adapter::Sidekiq}.
+      #
+      # @return [Hash<String, Object>]
+      def payload
+        @stripe_event.as_json
+      end
+
+      private
+
+      def method_missing(method_name, ...)
+        @stripe_event.send(method_name, ...)
+      end
+
+      def respond_to_missing?(...)
+        @stripe_event.respond_to?(...)
+      end
+    end
+  end
+end
--- a/README.md
+++ b/README.md
@@ -37,9 +37,92 @@ Bundle your dependencies and run the installation generator:
 bin/rails generate solidus_stripe:install
 ```
 
+### Webhooks
+
+This library makes use of some [Stripe
+webhooks](https://stripe.com/docs/webhooks).
+
+On development, you can [test them by using Stripe CLI](https://stripe.com/docs/webhooks/test).
+
+Before going to production, you'll need to [register the
+`/solidus_stripe/webhooks` endpoint with
+Stripe](https://stripe.com/docs/webhooks/go-live), and make sure to subscribe
+to the following events:
+
+[TBD]
+
+In both environments, you'll need to create a
+`solidus_stripe.webhook_endpoint_secret` credential with [the webhook signing
+secret](https://stripe.com/docs/webhooks/signatures):
+
+```bash
+# For development, add `--environment development`
+bin/rails credentials:edit
+```
+
+```yaml
+# config/credentials.yml.enc
+solidus_stripe:
+  webhook_endpoint_secret: "whsec_..."
+```
+
 ## Usage
 
-<!-- Explain how to use your extension once it's been installed. -->
+### Custom webhooks
+
+You can also use [Stripe webhooks](https://stripe.com/docs/webhooks) to trigger
+custom actions in your application.
+
+First, you need to register the event you want to listen to, both [in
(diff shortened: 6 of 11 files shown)


lib/solidus_stripe/configuration.rb
```diff
@@ -1,10 +1,24 @@
...
 module SolidusStripe
   class Configuration
-    # Define here the settings for this extension, e.g.:
-    #
-    # attr_accessor :my_setting
+    # @!attribute [rw] webhook_events
+    #  @return [Array<Symbol>] stripe events to handle. You also need to
+    #  register them in the Stripe dashboard. For an event `:foo`, a matching
+    #  `:"stripe.foo"` event will be registered in {Spree::Bus}.
+    attr_accessor :webhook_events
```
kennyadsl, line 11: Can user remove events that we need for the extension with this preference?

app/controllers/solidus_stripe/webhooks_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+module SolidusStripe
+  class WebhooksController < Spree::BaseController
+    SIGNATURE_HEADER = "HTTP_STRIPE_SIGNATURE"
+
+    skip_before_action :verify_authenticity_token, only: :create
+
+    respond_to :json
+
+    def create
+      request.body.read
```
kennyadsl: Thoughts on moving this to a service? I read your point that we are not ready with the infrastructure, and that makes sense. But when we will, and want to convert this, it will probably be easier because we can deprecate that class. 

No strong preference here: in the @elia's PR there are several controllers not using services, so this is more a general discussion about the approach we want to take.

config/routes.rb
```diff
@@ -2,4 +2,5 @@
 
 SolidusStripe::Engine.routes.draw do
   resources :payment_intents, only: :create
+  resources :webhooks, only: :create, format: false
```
kennyadsl, line 5: I think there could be conflicts here if we don't use some prefixes. `/webhooks` has high chances to be already taken.

kennyadsl (commented):
Left some comments but overall looks great!

lib/solidus_stripe/configuration.rb
waiting-for-dev: No, they can't, as we add them into the `SolidusStripe::Webhook::Event::CORE_EVENTS` constant.

config/routes.rb
waiting-for-dev: The generator will mount the routes within a `solidus_stripe` namespace, so I'll end up being `/solidus_stripe/webhooks`. Do you think that's enough? Or do you think people might end up mounting the engine on the root namespace?

kennyadsl: That's perfect. I was confused reading the PR description when you mention `POST /webhooks`. Solved!

elia: This should be scoped to `/solidus_stripe`, see https://github.com/solidusio/solidus_stripe/blob/bcbf03ecca1e1a179903db43e702aedf76b06e79/lib/generators/solidus_stripe/install/install_generator.rb#L20

lib/solidus_stripe/webhook/event.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      # TBD
+      CORE_EVENTS = Set[*%i[]].freeze
+      private_constant :CORE_EVENTS
+
+      # @api private
+      def self.register(user_events:, bus:, core_events: CORE_EVENTS)
+        (core_events + user_events).each do |event|
+          bus.register(:"stripe.#{event}")
+        end
```
elia: Do we really need to make a difference between core and users events? If we're going to copy the setup for "core" events to the target application they'll be all "user events" right?

The only thing I think could be core is the default implementation for some of the events, so that the user is free to adjust everything to the needs of the specific store.

Thoughts on removing the distinction?

lib/solidus_stripe/engine.rb
```diff
@@ -14,6 +14,16 @@ class Engine < Rails::Engine
       app.config.spree.payment_methods << 'SolidusStripe::PaymentMethod'
     end
 
+    initializer "solidus_stripe.pub_sub", after: "spree.core.pub_sub" do |app|
+      app.reloader.to_prepare do
+        require "solidus_stripe/webhook/event"
```
elia: This is not going to load the file but once, should it be moved outside of the `to_prepare` block?

app/controllers/solidus_stripe/webhooks_controller.rb
elia: Some of this could probably be kept inside `SolidusStripe::Webhook::Event`, something like:

```rb
def (SolidusStripe::Webhook::Event).from_request(request)
  tolerance = SolidusStripe.configuration.webhook_signature_tolerance
  secret = Rails.application.credentials.solidus_stripe[:webhook_endpoint_secret] # move to a configuration?
  signature_header = request.headers[SIGNATURE_HEADER]

  Stripe::Webhook.construct_event(payload, signature_header, secret, tolerance: tolerance)
end
```

app/controllers/solidus_stripe/webhooks_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+        .then { |payload| Stripe::Webhook.construct_event(payload, signature_header, secret, tolerance: tolerance) }
+        .then { |stripe_event| SolidusStripe::Webhook::Event.new(stripe_event: stripe_event) }
+        .then { |event| Spree::Bus.publish(event) }
```
elia: What happens when we receive an unknown event? 
Is that ignored by the bus or raises an exception?
Should we only publish registered events?

(please excuse my ignorance of the new bus on this topic)

elia (commented):
A few questions/sugestions

lib/solidus_stripe/engine.rb
waiting-for-dev: Good catch. Fixed.

lib/solidus_stripe/webhook/event.rb
waiting-for-dev: If everything is going to be handed to the user application, then we can remove the distinction. However, if we provide default implementations, that's not the case, right? 🤔 Maybe I'm missing something, but we probably need to capture incoming webhooks for payment capture so that our models are updated accordingly. That's not something I imagine belonging to the copied templates but to the core models. Am I wrong?

app/controllers/solidus_stripe/webhooks_controller.rb
waiting-for-dev: When an unknown event is received, an exception is raised, rendering the names of all registered events. That helps users be protected against typos. For instance, say that you subscribe by accident to an event `:"stripe.charge.succeded"` (there's a missing `e`). If we don't raise when there's an incoming `chage.succeeded` stripe event, users can have a hard time debugging.

More generally, getting an event you weren't anticipating is something unexpected, so I think an exception makes sense here.

What do you think?

app/controllers/solidus_stripe/webhooks_controller.rb
waiting-for-dev: Moving things to services requires some decisions to be made, like the interface we create for a result object. I think it doesn't pay off to build a skeleton here that will need to be deprecated later. At most, as @elia suggested, we can try to move some things into collaborators. The `from_request` method makes sense, but I'd also move the creation of the `SolidusStripe::Webhook::Event` there. This way, an instance of the collaborator will be involved. I'd mark the method as `@api private` so we're free to remove it later. Thoughts?

waiting-for-dev:
I'm currently investigating the issue on CI on Ruby 2.7.

When doing:

```ruby
        .then { |stripe_event| SolidusStripe::Webhook::Event.new(stripe_event: stripe_event) }
        .then { |event| Spree::Bus.publish(event) }
```

The event yield on the second step is a legit `SolidusStripe::Webhook::Event`. However, `Spree::Bus#publish` receives a `Hash`. For now, I've narrowed down the issue to any object responding to `#to_hash`. `Stripe::Event` implements it, and `SolidusStripe::Webhook::Event` delegates to it.

elia:
@waiting-for-dev would probably be wiser to use `#to_h` which is the generic conversion form, `#to_hash` `#to_str` `#to_ary` usually are for objects that are intrinsically of the type you're converting to. E.g. a symbol has `#to_s` and `#to_sym` but not `#to_str`.

waiting-for-dev:
> @waiting-for-dev would probably be wiser to use `#to_h` which is the generic conversion form, `#to_hash` `#to_str` `#to_ary` usually are for objects that are intrinsically of the type you're converting to. E.g. a symbol has `#to_s` and `#to_sym` but not `#to_str`.

@elia, the `#to_hash` method is coming from the `Stripe::Event` object, which we don't control. `SolidusStripe::Webhook::Event` is simply delegating to it

elia:
> @elia, the #to_hash method is coming from the Stripe::Event object, which we don't control. SolidusStripe::Webhook::Event is simply delegating to it

Yes, I meant that's the root cause in my opinion, combined with 2.7 handling of keyword-arguments splatting and potentially the `then` implementation.

waiting-for-dev:
> > @elia, the #to_hash method is coming from the Stripe::Event object, which we don't control. SolidusStripe::Webhook::Event is simply delegating to it
> 
> Yes, I meant that's the root cause in my opinion, combined with 2.7 handling of keyword-arguments splatting and potentially the `then` implementation.

Thanks, @elia, it turned out to be related to an override in Solidus itself. See https://github.com/solidusio/solidus/pull/4875

app/controllers/solidus_stripe/webhooks_controller.rb
elia: I see the point, which brings me to another question. I thought that stripe sends out all events it can, or do we have control in the dashboard over what events we want to receive?

I'd say if we have control then it's fine to raise; otherwise we risk having a bunch of exceptions from all the tens of events sent by stripe: in that case maybe we can filter the list of what we want to forward.

app/controllers/solidus_stripe/webhooks_controller.rb
elia: I think I need to see the code in order to understand fully the proposal, but I'm down with it from what I understand… and it's way easier to go from `@api private` to `@api public` than the other way around 🙌 

The main goal is to have building block that are easy to reuse in case the target app wants (or needs) to customize or redo the controller.

lib/solidus_stripe/webhook/event.rb
elia: If possible I would err on the side of allowing the target app to take control of core events. Stripe has many features and maybe an additional fraud check or external system integration needs to be added.

The implementations we provide out of the box should be more examples and (I think) allow breaking them apart leaving the glue code in the target application. That to me would be the best of both worlds.

I'm assuming it's possible to do that, but if not maybe the current division is already the best we can do

waiting-for-dev: As you can create more than one subscriber for a given event, we're good at letting users add new stuff.

As per our subscribers, we should design them with the SRP in mind. For instance, if we need to do two things on `payment.succeeded`, let's create two subscribers for the same event and not a single one. This way, we pinpoint the need for any of the regular Solidus approaches to modify core behavior. On top of that,in this case, Omnes allows you to [unsubscribe](https://github.com/nebulab/omnes#unsubscribing--clearing) listeners from the bus 👍

app/controllers/solidus_stripe/webhooks_controller.rb
waiting-for-dev: My understanding is that stripe requires you to [register the events you're interested in on the dashboard](https://stripe.com/docs/webhooks/go-live#configure-webhook-settings).

waiting-for-dev: A possible pain point would be for users listening to all events:


However, Omnes is designed with safety in mind, so it doesn't allow publishing events if they haven't been registered before. We could make some changes upstream to introduce namespace events and different strategies for checking the event name. For instance, allowing any event starting with `stripe.` if the user prefers.

I'd say we should move forward now and work on it in a later stage if we think or find out that's a valid concern. Thoughts?

elia: The fact that Stripe lets you list the events you wanna receive settles it for me, we just need to instruct to list only the ones supported out-of-the-box when installing (or provide the code needed to do it from the API) and that's it 🙌 

Thanks for explaining 👍

app/controllers/solidus_stripe/webhooks_controller.rb
waiting-for-dev: Please, take a look now. I moved the logic other than publishing the event to a class method in `SolidusStripe::Webhook::Event`.

waiting-for-dev:
@elia, @kennyadsl, I have addressed your suggestions. Please, could you take another look? I also added the new settings to the initializer template (I had forgotten) and refactored the test fixture to easily share them.

kennyadsl (approved):
👍 Thanks @waiting-for-dev

elia (approved):
Awesome!

I tried to understand what's going on with the sqlite failure in the CI but couldn't from the diff, any ideas @waiting-for-dev?

waiting-for-dev:
> Awesome!
> 
> I tried to understand what's going on with the sqlite failure in the CI but couldn't from the diff, any ideas @waiting-for-dev?

We already talked about it offline, but for context this is the issue I talked in https://github.com/solidusio/solidus_stripe/pull/157#issuecomment-1403512955

PR #158, opened 2023-02-06
Move payment intent id tracking from payment source to payment

### LINKED ISSUE #154: Link Stripe Payment Intents to `Spree::Payment`
## What

We need to map each `Spree::Payment` to a single Payment Intent, and thus have a single Stripe `Spree::Payment` for a given order that will track all the payment attempts made for the Payment Intent. Stripe recommends having a single payment intent per order.

```[tasklist]
### Tasks
- [x] connect the payment intent id to the payment transaction_id
- [x] implement the state mapping as described below
```

### Dependencies

- blocks #153 

### Mapping between Solidus and Stripe payment states

The following is the tentative mapping between the states in the two systems, since in Stripe the "failed" state is just going back to the starting point we'll treat anything before a final state as `:processing`.

image

## Why

Since Stripe considers updates sent via webhooks to be the most reliable source of truth, we need a place in which to track the history of each payment tracking _attempts_, failures, refusals, successes etc.

The entity that can best take up this role seems to be `Spree::Payment`.

### LINKED ISSUE #167: Ensure we're running the CI without leaking test credentials
## What

We're going to run the CI using our test credentials, those should not be leaked to forks or be accessible in any way, including VCR cassettes, logs, etc.

## Why

Even if they're test credentials they're still tied to a real account and not public domain. We're not aware of any alternative way for live testing.

After the initial release we'll probably try to replace things with VCR, stripe-mock, or something else, assuming it will still allow an end to end test for system specs.

elia:
## Summary

- Fixes #154 
- Fixes #167 

## Checklist


app/controllers/solidus_stripe/intents_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+require 'stripe'
+
+class SolidusStripe::IntentsController < Spree::BaseController
+  include Spree::Core::ControllerHelpers::Order
+
+  before_action :load_payment_method
+
+  def payment_confirmation
+    intent = @payment_method.gateway.request { Stripe::PaymentIntent.retrieve(id: params[:payment_intent]) }
+    payment = @payment_method.find_in_progress_payment_for(current_order)
```
waiting-for-dev, line 11: Are we ensuring that only one in-progress payment exists at a given point for an order & Stripe payment method?

app/controllers/solidus_stripe/intents_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      return
+    end
+
+    current_order.state = :payment
+    # https://stripe.com/docs/payments/intents?intent=payment
+    case intent.status
+    when 'requires_payment_method'
+      # this state after an attempt at processing a payment means that it failed.
+      payment.failure! unless payment.failure?
+      ensure_state_is(payment, :processing)
```
waiting-for-dev: How is it that we change the state to `:failure` but later expect it to be `:processing`?

app/controllers/solidus_stripe/intents_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+    when 'requires_confirmation'
+      ensure_state_is(payment, :processing)
```
waiting-for-dev: Aren't those `ensure_state_is` defensive? Shouldn't our logic ensure that the state is the expected one?

app/controllers/solidus_stripe/intents_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+class SolidusStripe::IntentsController < Spree::BaseController
+  include Spree::Core::ControllerHelpers::Order
+
+  before_action :load_payment_method
+
+  def payment_confirmation
+    intent = @payment_method.gateway.request { Stripe::PaymentIntent.retrieve(id: params[:payment_intent]) }
+    payment = @payment_method.find_in_progress_payment_for(current_order)
+
+    unless %w[confirm payment].include?(current_order.state.to_s)
```
waiting-for-dev, line 18: Is there any scenario where that could be false?

app/controllers/solidus_stripe/intents_controller.rb
waiting-for-dev: I guess it's related to:

> The most important difference is that on Stripe a failure is not a final state, rather just a way to start over.

However, that's something we're not handling yet, right?

app/controllers/solidus_stripe/intents_controller.rb
elia: Only by using a single entry point (which is the method above) but we surely can do better

app/controllers/solidus_stripe/intents_controller.rb
elia: I guess you can experiment with checking out from a product page and things like that, or mess with an order using multiple tabs. But as you pointed out below it's mostly defensive, in order to assert expectations and prerequisites.

app/controllers/solidus_stripe/intents_controller.rb
elia: Good call! The expectation got it right, and the call to `failure!` is legacy of what I tried before understanding that caveat. But since that part is not covered by test yet it got through, fixing it 👍

app/controllers/solidus_stripe/intents_controller.rb
elia: Yes they are, and that's because of how many moving parts and configuration options we have in Solidus. Eventually when we'll have better coverage we might want to relax that a bit. 

But they're also very lightweight so maybe the work as runnable documentation? Anyway I think we can decide later

elia (commented):
I tried to answer your questions, knowing that block of code is pretty raw and hoping it will improve with the extraction 😊

app/models/solidus_stripe/gateway.rb
```diff
@@ -19,15 +19,6 @@ def initialize(options)
 
     attr_reader :client
 
-    # Authorizes a certain amount on the provided payment source.
-    #
-    # @see #purchase
-    def authorize(amount_in_cents, source, options = {})
```
kennyadsl, line 34: Does this mean auto capture is always on?

lib/generators/solidus_stripe/install/templates/app/assets/stylesheets/spree/frontend/solidus_stripe.css
```diff
@@ -6,3 +6,8 @@ the installer will append this file to the app vendored assets here: 'vendor/ass
 .solidus-stripe-payment {
   padding: 1em 0;
 }
+
```
kennyadsl: Does `Dim disabled submit buttons` need a rewrite?

lib/generators/solidus_stripe/install/templates/config/initializers/solidus_stripe.rb
```diff
@@ -13,6 +13,29 @@
...
+#
+# We're providing a inline module to make it easier to remove the mokey-patch
+# once the upstream PR is merged and generally available.
+#
+# Fixes https://github.com/solidusio/solidus/issues/2680
+# Fixed by https://github.com/solidusio/solidus/pull/4909
+begin
+  module SolidusStripe::PaymentCreateWithExistingPayment
+    def build
+      @payment ||= order.payments.find(attributes[:id]) if attributes[:id]
```
kennyadsl: With this whole picture in mind, I had some thoughts about how we can upstream this to core. What about marking some payment methods as "accepting multiple attempts", maybe with some overridable methods similar to the existing `support_profiles?`. When it's `true` (and Stripe will be) core will act differently taking the already existing payment of the same type, instead of passing the id explicitly, which opens the door to extra security implications.

kennyadsl (commented):
Left some comments!

app/models/solidus_stripe/gateway.rb
elia: No, it just means that the authorization happens in a different way, through the frontend and the payment_method helper to find-or-create the current in-progress payment.

lib/generators/solidus_stripe/install/templates/app/assets/stylesheets/spree/frontend/solidus_stripe.css
elia: It should be fine, we're still dimming disabled submit-buttons, maybe you would like to see a reference to the cursor change?

lib/generators/solidus_stripe/install/templates/config/initializers/solidus_stripe.rb
elia: > extra security implications.

I tried to think this through and the only issue I can find is that [`amount` is a permitted parameter](https://github.com/solidusio/solidus/blob/master/core/lib/spree/permitted_attributes.rb#L72), potentially allowing amount updates. Otherwise I think the fetching of this payment is pretty tight, only allowing payments already attached to the order. I'll try to work around that and update the PR.

Is there any other security issue we can address or is it unknown way in which this could be abused?

> What about marking some payment methods as "accepting multiple attempts", maybe with some overridable methods similar to the existing support_profiles?. When it's true (and Stripe will be) core will act differently taking the already existing payment of the same type

I'm a bit disaffected with the current flow, if possible I'd like to simplify things and reduce coupling between the legacy active-merchant/credit-card support and payments so that we (and everyone else) can eventually swap it with a better implementation.

spec/system/frontend/solidus_stripe/checkout_spec.rb
```diff
@@ -0,0 +1,141 @@
+# frozen_string_literal: true
+
+require 'solidus_stripe_spec_helper'
+
+RSpec.describe "Checkout with Stripe", :vcr, :js do
```
waiting-for-dev: Looks like we're not using `vcr`?

waiting-for-dev:
@elia, I've found that when there's Stripe, the other available payment methods are not rendered to the user. Is that expected?

app/controllers/solidus_stripe/intents_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+    when 'requires_confirmation', 'requires_action', 'processing'
+      ensure_state_is(payment, :processing)
+      current_order.next!
+    when 'requires_capture'
+      payment.pend! unless payment.pending?
+      current_order.next!
+      ensure_state_is(current_order, :confirm)
+      ensure_state_is(payment, :pending)
+    when 'succeeded'
+      payment.completed! unless payment.completed?
```
waiting-for-dev, line 45: @elia, after local testing, I've found that this case is not covered by tests or manual checkout. Am I missing something?

elia: It's missing testing with autocapture on

elia:
> @elia, I've found that when there's Stripe, the other available payment methods are not rendered to the user. Is that expected?

@waiting-for-dev It's solidus giving priority to methods attached to the current Spree::Store, I think solidus master will add it by default through the admin but maybe not from samples

app/models/solidus_stripe/payment_method.rb
```diff
@@ -4,40 +4,133 @@ module SolidusStripe
...
-          payment_intent.save
+      def create_in_progress_payment_for(order)
+        intent = gateway.request do
+          Stripe::PaymentIntent.create({
+            amount: gateway.to_stripe_amount(
+              order.display_total.money.fractional,
+              order.currency,
+            ),
+            currency: order.currency,
+            capture_method: 'manual',
```
waiting-for-dev: I understand that here we're telling Stripe that we want to capture payments manually. However, shouldn't we integrate that into Solidus logic for auto-capture?

https://github.com/solidusio/solidus/blob/7be1937c16c9c8bf689010a91e762a833257de87/core/app/models/spree/payment_method.rb#L157

elia: Thanks, I set it as a placeholder initially and forgot to connect it to the query method 👍

spec/system/frontend/solidus_stripe/checkout_spec.rb
elia: Removed thanks!

lib/generators/solidus_stripe/install/templates/app/assets/stylesheets/spree/frontend/solidus_stripe.css
kennyadsl: Lol, I thought `Dim` was some sort of leftover vim shortcut that went wrong. 🤦, also because I know you don't use Vim!

elia: If I was to use vim you would see those weird extra characters all over the place! ⌨️ 🤣

lib/generators/solidus_stripe/install/templates/config/initializers/solidus_stripe.rb
waiting-for-dev: > I tried to think this through and the only issue I can find is that [amount is a permitted parameter](https://github.com/solidusio/solidus/blob/master/core/lib/spree/permitted_attributes.rb#L72), potentially allowing amount updates

That's serious enough 🙂

app/controllers/solidus_stripe/intents_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      raise "The payment intent id doesn't match"
+    end
+
+    unless %w[confirm payment].include?(current_order.state.to_s)
+      redirect_to main_app.checkout_state_path(current_order.state)
+      return
+    end
+
+    current_order.state = :payment
+    payment.log_entries.create!(details: { payment_intent: intent.to_json })
```
waiting-for-dev: Should we add the class to [this preference](https://github.com/solidusio/solidus/blob/7be1937c16c9c8bf689010a91e762a833257de87/core/lib/spree/app_configuration.rb#L185)?

app/models/solidus_stripe/payment_method.rb
```diff
@@ -4,40 +4,133 @@ module SolidusStripe
...
+      def actions
+        %w[capture void credit]
+      end
 
-    alias cart_partial_name partial_name
-    alias product_page_partial_name partial_name
+      # @param payment [Spree::Payment] the payment we want to know if can be captured
+      # @return [Boolean] true when the payment is in the pending or checkout states
+      def can_capture?(payment)
+        payment.pending? || payment.checkout?
```
waiting-for-dev: I don't understand why a payment that is still on the checkout step can be captured 🤔

app/models/solidus_stripe/payment_method.rb
```diff
@@ -4,40 +4,133 @@ module SolidusStripe
...
+          .where.not(state: %w[completed invalid void]) # in_progress
+          .find_by(payment_method: self)
+      end
 
-      if payment_intent && payment_intent.customer.blank?
-        payment.payment_method.gateway.request do
-          payment_intent.customer = Stripe::Customer.new email: payment.order.email
-          payment_intent.customer.save
-          payment_intent.save
+      def create_in_progress_payment_for(order)
```
waiting-for-dev, line 69: What about checking here whether `#find_in_progress_payment_for` already returns a payment?

app/controllers/solidus_stripe/intents_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    # https://stripe.com/docs/payments/intents?intent=payment
+    case intent.status
```
waiting-for-dev, line 32: As we agreed on handling everything with the webhooks, what about making this PR more straightforward by removing all the logic from here and simply displaying a message? That would also play better with the current test coverage.

waiting-for-dev:
@elia, while testing my work, I found a way to create two stripe payments for the same order. Not sure if that's exploitable, but we should ensure only one is possible, at least for the initial release.

[two_payments](https://user-images.githubusercontent.com/52650/220097307-73b64c74-4483-4075-b254-a612a8613744.mp4)

elia:
@waiting-for-dev good catch, I'll move that into a different issue as we should probably properly cover it with tests and whatnot. Btw I'm working on avoiding the use of the `:id` attribute.

waiting-for-dev:
@elia, here are a few things I noticed while testing:

- A Stripe payment on the `:processing` status is created when reaching the payment step. That's it: before selecting Stripe as the payment method or entering any payment information. At that moment, we're also receiving the `payment_intent.created` webhook event. That means that Stripe can't be used along with other payment methods. Besides, creating payments for orders that could not go through the payment step is very weird.
- After filling in the payment info, we get the `payment_intent.amount_capturable_updated` webhook event. That's how Stripe wants to communicate that the [payment is ready to be captured](https://stripe.com/docs/payments/payment-intents/verifying-status#webhooks). However, that's before the user confirms that they want to place the order.
screenshot-stripe com-2023 02 21-09_56_11
- We don't get any webhook event once the user confirms the order.
- Minor issue: we're not collecting billing information.
- Minor issue: before confirming the order, the Stripe payment is on the `:pending` state instead of `:checkout` as I saw the built-in Solidus payment method do.

Would it be possible to:
- Defer the payment creation until after the payment information has been created.
- Defer the payment intent creation until after the payment information has been created.
- Defer receiving the `amount_capturable_updated` event (I haven't checked what it's triggering it, yet) until after the order has been confirmed.

Of course, I'm more than happy to help with anything needed.

waiting-for-dev:
Updating with the conclusions after offline discussion:

> * A Stripe payment on the `:processing` status is created when reaching the payment step. That's it: before selecting Stripe as the payment method or entering any payment information. At that moment, we're also receiving the `payment_intent.created` webhook event. That means that Stripe can't be used along with other payment methods. Besides, creating payments for orders that could not go through the payment step is very weird.

**Reason:**
We need the Payment Intent to be able to render Stripe's Payment Element. We're currently mapping one-to-one payment intent and payment, so we need the latter.

**Solution:**
If possible, decouple payment intent from payment. Create a communication layer between Solidus and Stripe, and only update Solidus' payment at the right moment.

**Workaround:**
@elia pointed out that those dangling payments will eventually be voided if another payment method is used. @elia, do we need to confirm that, or do we have a reference for that behavior?

> * After filling in the payment info, we get the `payment_intent.amount_capturable_updated` webhook event. That's how Stripe wants to communicate that the [payment is ready to be captured](https://stripe.com/docs/payments/payment-intents/verifying-status#webhooks). However, that's before the user confirms that they want to place the order.

**Reason:**
We need to [confirm the payment](https://github.com/solidusio/solidus_stripe/blob/ad94991ed6d715356686f43a5494ce7114d3649b/lib/generators/solidus_stripe/install/templates/app/javascript/controllers/solidus_stripe_payment_controller.js#L51) while the payment element is still present. However, Solidus checkout flow will still have one step left after submitting the payment (`confirm` -> `complete`). As that's another request, we've lost track of the payment element by then.

**Solution:**
- We could use [server-side payment confirmation](https://stripe.com/docs/payments/defer-intent-creation-confirm-server), but that's a Beta feature. A "next action" response field would tell us what we need to do from the JS part.
- We could remove the `confirm` step from the Solidus' checkout flow.

**Workaround:**:
1 - Force `auto_capture` to always be off for Stripe.
2 - Warn users that `solidus_stripe` will authorize payments before the order is completed. Suggest using a background job to cancel dangling payments.

> * We don't get any webhook event once the user confirms the order.

**Reason:**
Same as above.

**Solution:**
Same as above.

**Workaround:**
We should store the webhook and act on it at a later stage.

> Minor issue: before confirming the order, the Stripe payment is on the :pending state instead of :checkout as I saw the built-in Solidus payment method do.

**Reason:**
Same as above. After payment but before confirming the order, the Stripe payment has been confirmed and shouldn't have been captured, so `:pending` is the right state at this point.

waiting-for-dev:
> * Minor issue: we're not collecting billing information.

I created https://github.com/solidusio/solidus_stripe/issues/175 for this one

elia:
@waiting-for-dev thanks for jotting down the notes from the offline discussion, I only have one "correction" on the first point:

> Solution:
> If possible, decouple payment intent from payment. Create a communication layer between Solidus and Stripe, and only update Solidus' payment at the right moment.

I'd like to challenge this, what would be the upside of doing this?
In the current state of things I only see the downsides of having a more complex schema and relations to manage properly, additional records in the DB, additional UI in the admin, etc.

maybe we can just call it a "Potential Solution"? Thoughts?

waiting-for-dev:
> I'd like to challenge this, what would be the upside of doing this? In the current state of things I only see the downsides of having a more complex schema and relations to manage properly, additional records in the DB, additional UI in the admin, etc.
> 
> maybe we can just call it a "Potential Solution"? Thoughts?

The upside is a cleaner interface without undesired side effects (the `:processing` payment). But, of course, all of those are only potential solutions, and if we decide to tackle them at some point, we'll need to evaluate all the trade-offs 🙂

elia:
@waiting-for-dev I extracted the auto_capture fix to this issue https://github.com/solidusio/solidus_stripe/issues/178

README.md
```diff
@@ -128,6 +128,42 @@ SolidusStripe.configure do |config|
...
+the agreement to the terms of service and an order summary in the payment step and apply this patch:
+
+```patch
+--- a/templates/app/controllers/checkouts_controller.rb
++++ b/templates/app/controllers/checkouts_controller.rb
+@@ -48,6 +48,8 @@ def redirect_on_failure
+   end
+
+   def transition_forward
++    @order.next if @order.has_checkout_step?("payment") && @order.payment?
```
waiting-for-dev, line 164: I tried a similar hack but in `solidus-core` itself (calling `.next!` after the transition to `:confirm`), and there were a bunch of errors:

```
rspec ./spec/models/spree/promotion_code_spec.rb:237 # Spree::PromotionCode completing multiple orders with the same code resets the state of the order
rspec ./spec/models/spree/promotion_code_spec.rb:231 # Spree::PromotionCode completing multiple orders with the same code increases the total to remove the promo
rspec ./spec/models/spree/promotion_code_spec.rb:225 # Spree::PromotionCode completing multiple orders with the same code adjusts the promo_total
rspec ./spec/models/spree/promotion_code_spec.rb:219 # Spree::PromotionCode completing multiple orders with the same code makes the promotion ineligible
rspec ./spec/models/spree/reimbursement_spec.rb:112 # Spree::Reimbursement#perform! refunds the total amount
rspec ./spec/models/spree/reimbursement_spec.rb:117 # Spree::Reimbursement#perform! creates a refund
rspec ./spec/models/spree/reimbursement_spec.rb:172 # Spree::Reimbursement#perform! triggers the reimbursement mailer to be sent via subscribed event
rspec ./spec/models/spree/reimbursement_spec.rb:142 # Spree::Reimbursement#perform! with included tax saves the included tax and refunds the total
rspec ./spec/models/spree/reimbursement_spec.rb:157 # Spree::Reimbursement#perform! when reimbursement cannot be fully performed does not send a reimbursement email and raises IncompleteReimbursement error
rspec ./spec/models/spree/reimbursement_spec.rb:127 # Spree::Reimbursement#perform! with additional tax saves the additional tax and refunds the total
rspec ./spec/models/spree/reimbursement_spec.rb:166 # Spree::Reimbursement#perform! when exchange is required generates an exchange shipment for the order for the exchange items
rspec ./spec/models/spree/order/checkout_spec.rb:417 # Spree::Order with default state machine from payment with confirmation required transitions to confirm
rspec ./spec/models/spree/order/checkout_spec.rb:430 # Spree::Order with default state machine from payment when payment is not required does not call process payments
rspec ./spec/models/spree/order/checkout_spec.rb:291 # Spree::Order with default state machine from delivery attempts to apply free shipping promotions
rspec ./spec/models/spree/order/checkout_spec.rb:323 # Spree::Order with default state machine from delivery without payment required transitions to complete
rspec ./spec/models/spree/order/checkout_spec.rb:358 # Spree::Order with default state machine from delivery correctly determining payment required based on shipping information with a shipment that is free skips payment, transitions to confirm
```

I didn't look into the details, and probably at least some failures are because of the test setup itself.

waiting-for-dev (commented):
Thanks, @elia. After the conversation yesterday, I have all the tradeoffs clear in mind. There are a few things we still need to figure out, though. Besides some minor comments I made in past reviews, the bigger problem I'm experiencing is that while testing, I'm receiving not only the `payment_intent.amount_capturable` event (authorization), but also the `charge.succeeded` one (actual money transaction).

Besides, I think we should remove the handling from the callback controller. Some of the paths are not tested, plus I'm not sure some of the business logic is accurate (as the Solidus checkout flow will already take some of the steps by itself). As we agreed on do the handling on webhooks, I think the conversation we'll be easier if we do that on that PRs. Thoughts?

README.md
waiting-for-dev: Not saying to remove the warning from the README, though.

waiting-for-dev:
> @elia, while testing my work, I found a way to create two stripe payments for the same order. Not sure if that's exploitable, but we should ensure only one is possible, at least for the initial release.

Confirmed that's no longer an issue. After the second payment, the first one gets canceled (and reimbursed for now).

app/models/solidus_stripe/payment_method.rb
```diff
@@ -4,40 +4,153 @@ module SolidusStripe
...
+          intent = gateway.request do
+            Stripe::PaymentIntent.create({
+              amount: gateway.to_stripe_amount(
+                order.display_total.money.fractional,
+                order.currency,
+              ),
+              currency: order.currency,
+
+              # The capture method should stay manual in order to
+              # avoid capturing the money befor the order is completed.
```
waiting-for-dev:
```suggestion
              # avoid capturing the money before the order is completed.
```

waiting-for-dev:
> the bigger problem I'm experiencing is that while testing, I'm receiving not only the `payment_intent.amount_capturable` event (authorization), but also the `charge.succeeded` one (actual money transaction).

As you suggested offline, the payment is not captured 👍 It's confusing that stripe sends a `charge.succeeded` webhook, but its `amount_captured` field is 0, and it doesn't change the `requires_capture` status. See [example](https://dashboard.stripe.com/test/payments/pi_3MeFYQJvEPu9yc7w1xHDhpVM).

elia:
> Besides, I think we should remove the handling from the callback controller. Some of the paths are not tested, plus I'm not sure some of the business logic is accurate (as the Solidus checkout flow will already take some of the steps by itself). As we agreed on do the handling on webhooks, I think the conversation we'll be easier if we do that on that PRs. Thoughts?

@waiting-for-dev Would it be fine if we leave it there as a placeholder and it gets removed when we have a replacement through webhooks? To me it was very helpful for being able to complete payments and test the full flow while we lack webhooks

waiting-for-dev:
> @waiting-for-dev Would it be fine if we leave it there as a placeholder and it gets removed when we have a replacement through webhooks? To me it was very helpful for being able to complete payments and test the full flow while we lack webhooks

Sure thing. I'll create an issue for that later.

waiting-for-dev:
@elia, it looks there's a problem when trying to show a Stripe payment from the admin panel.

[screencast-localhost_3000-2023.02.22-13_05_37.webm](https://user-images.githubusercontent.com/52650/220615668-db4bc8a6-ec80-47fa-81d9-07dd706b4835.mp4)

elia:
@waiting-for-dev I'm fixing that in another PR, I can backport the fix or keep it there (we also need to add specs for those operations I think), otherwise we can delay the fix to the next PR, thoughts?

lib/generators/solidus_stripe/install/templates/config/initializers/solidus_stripe.rb
elia: This should be now addressed, the PaymentCreate patch will return early if the `:id` parameter is provided, so no update of any sort will happen.

app/controllers/solidus_stripe/intents_controller.rb
elia: Would mean to add a Hash, but as you pointed out in a more recent comment then the admin UI blows up anyway as it expects an active merchant response

app/models/solidus_stripe/payment_method.rb
elia: TBH I have no idea, that's code that was copied from Spree::PaymentSource to the payment method in order for it to act as a source, so we could have payments without sources.

app/models/solidus_stripe/payment_method.rb
elia: I'm trying to follow the rails style of having methods explicitly named  `find_or_create_` that internally call `find_… || create_…`. Do you see any advantage in moving the check in here?

app/controllers/solidus_stripe/intents_controller.rb
elia: I'd like to keep the complete checkout flow to be working with proper payment status updates, and until we have webhooks this has served this function well. It's surely intended as nothing more than a draft of what the webhook response will be.

Thoughts on leaving it as it is or at least keeping what's required for the checkout spec to keep functioning?

app/models/solidus_stripe/payment_method.rb
waiting-for-dev: I saw that you're tackling that in `find_in_progress_payment_for` with:

```ruby
if payments.size > 1
  payments[1...-1].each(&:invalidate!)
end
```

app/controllers/solidus_stripe/intents_controller.rb
waiting-for-dev: 👍 Created #190

waiting-for-dev:
> @waiting-for-dev I'm fixing that in another PR, I can backport the fix or keep it there (we also need to add specs for those operations I think), otherwise we can delay the fix to the next PR, thoughts?

👍 I created #191

.circleci/config.yml
```diff
@@ -85,6 +85,7 @@ workflows:
     jobs:
       - run-specs:
           name: 'run-specs-ruby-<<matrix.ruby>>-db-<<matrix.database>>'
+          context: stripe-test-credentials
```
waiting-for-dev: We'll probably have unauthorized builds for contributions, as we experienced when configuring Slack alerts 😢 Not for now, but maybe we can use vcr in the future?

README.md
```diff
@@ -128,6 +128,42 @@ SolidusStripe.configure do |config|
...
+
+Reference: https://stripe.com/docs/payments/intents?intent=payment
+
+![image](https://user-images.githubusercontent.com/1051/217322027-f49081f5-0795-49f4-994e-285a9de5347c.png)
+
+### ⚠️ Warning: Authorization happens before the order is completed
+
+This setup implies that after the payment information are submitted to Stripe the payment is authorized, although
+the order is not yet completed. This can become an issue if a customers abandons the checkout at the `confirm`
+step and no action is taken to free up the money on the backend.
```
waiting-for-dev:
```suggestion
This setup implies a payment is authorized after the payment information is submitted to Stripe, although the order still needs to be completed. That can become an issue if a customer abandons the checkout at the `confirm` step and no action is taken to free up the money on the backend.
```

I found it hard to read.

README.md
```diff
@@ -128,6 +128,42 @@ SolidusStripe.configure do |config|
...
+
+In order to mitigate this issue we suggest to adapt the frontend merging the confirm and payment steps. Embedding
+the agreement to the terms of service and an order summary in the payment step and apply this patch:
```
waiting-for-dev:
```suggestion
We suggest adapting the frontend merging the confirm and payment steps to mitigate this issue: embed the agreement to the terms of service and an order summary in the payment step and apply this patch.
```

Same here 🙂

README.md
```diff
@@ -128,6 +128,42 @@ SolidusStripe.configure do |config|
...
+   end
+
+   def transition_forward
++    @order.next if @order.has_checkout_step?("payment") && @order.payment?
++
+     if @order.can_complete?
+       @order.complete
+     else
+```
+
```
waiting-for-dev, line 170: Considering that there might be some unexpected problems when merging both checkout steps, what do you think if we propose creating a background job as another option?

app/models/solidus_stripe/gateway.rb
```diff
@@ -43,11 +34,19 @@ def authorize(amount_in_cents, source, options = {})
...
-    def capture(_amount_in_cents, _transaction_id, options = {})
-      payment = options[:originator] or raise ArgumentError, "please provide a payment with the :originator option"
-      payment_intent_id = payment.source.stripe_payment_intent_id
+    def capture(amount_in_cents, payment_intent_id, options = {})
+      payment = options[:originator] or
+        raise ArgumentError, "please provide a payment with the :originator option"
 
-      raise ArgumentError, "missing transaction_id" unless payment_intent_id
+      unless payment_intent_id
+        raise ArgumentError, "missing transaction_id"
```
waiting-for-dev:
```suggestion
        raise ArgumentError, "missing payment_intent_id"
```

app/models/solidus_stripe/gateway.rb
```diff
@@ -43,11 +34,19 @@ def authorize(amount_in_cents, source, options = {})
     # @option options [Spree::Payment] :originator the payment on which the #capture is being performed (required)
     #
     # @return ActiveMerchant::Billing::Response
-    def capture(_amount_in_cents, _transaction_id, options = {})
-      payment = options[:originator] or raise ArgumentError, "please provide a payment with the :originator option"
-      payment_intent_id = payment.source.stripe_payment_intent_id
+    def capture(amount_in_cents, payment_intent_id, options = {})
+      payment = options[:originator] or
```
waiting-for-dev, line 38: What's the point of having `payment_intent_id` as an argument if we already have access to it through `options[:originator].payment_intent_id`?

app/models/solidus_stripe/gateway.rb
```diff
@@ -135,12 +137,9 @@ def void(_transaction_id, source, _options = {})
     #   being performed (required if source is nil)
     #
     # @return ActiveMerchant::Billing::Response
-    def credit(amount_in_cents, _source, _transaction_id, options = {})
-      refund = options[:originator]
-      payment = refund.payment
-      source = payment.source
+    def credit(amount_in_cents, _source, payment_intent_id, options = {})
```
waiting-for-dev, line 113: Same here, should we just ignore the `payment_intent_id` argument?

app/models/solidus_stripe/payment_method.rb
```diff
@@ -4,40 +4,153 @@ module SolidusStripe
   class PaymentMethod < ::Spree::PaymentMethod
     preference :api_key, :string
     preference :publishable_key, :string
+    preference :webhook_endpoint_secret, :string, default: ''
```
waiting-for-dev: Unless I'm missing something that's not used anywhere yet. https://github.com/solidusio/solidus_stripe/issues/161 will handle it, so I think we should remove this line for now.

app/models/solidus_stripe/payment_method.rb
waiting-for-dev: I think we can remove it, as the payment will go straightaway to the `:pending` state.

app/models/solidus_stripe/payment_method.rb
```diff
@@ -4,40 +4,153 @@ module SolidusStripe
...
+      def find_in_progress_payment_for(order)
+        payments = order.payments
+          .where.not(state: %w[completed invalid void]) # in_progress
+          .where(payment_method: self)
+          .order(:created_at)
+          .entries
+
+        if payments.size > 1
+          payments[1...-1].each(&:invalidate!)
+        end
```
waiting-for-dev: I'd like to see a test for that, as that's a security concern.

app/models/solidus_stripe/payment_method.rb
```diff
@@ -4,40 +4,153 @@ module SolidusStripe
...
+      alias product_page_partial_name partial_name
+      alias risky_partial_name partial_name
     end
 
-    def create_profile(payment)
-      payment_intent = payment.source.payment_intent
+    concerning :Order do
+      def find_in_progress_payment_for(order)
+        payments = order.payments
+          .where.not(state: %w[completed invalid void]) # in_progress
```
waiting-for-dev, line 53: Why not `.where(state: "pending")`?

app/models/solidus_stripe/payment_method.rb
```diff
@@ -4,40 +4,153 @@ module SolidusStripe
...
+          .entries
+
+        if payments.size > 1
+          payments[1...-1].each(&:invalidate!)
+        end
+
+        payment = payments.last
+
+        if payment && payment.amount != order.total
+          payment.checkout? ? payment.invalidate! : payment.cancel!
```
waiting-for-dev: I think we'll never be in the `:checkout` state.

lib/generators/solidus_stripe/install/templates/config/initializers/solidus_stripe.rb
```diff
@@ -13,6 +13,32 @@
...
+# We're providing a inline module to make it easier to remove the mokey-patch
+# once the upstream PR is merged and generally available.
+#
+# Fixes https://github.com/solidusio/solidus/issues/2680
+# Fixed by https://github.com/solidusio/solidus/pull/4909
+begin
+  module SolidusStripe::PaymentCreateWithExistingPayment
+    def build
+      if attributes[:id]
+        @payment = order.payments.find(attributes[:id])
```
waiting-for-dev: It looks like we're still providing the id of an existing payment?

waiting-for-dev (commented):
Left more comments, @elia. Everything should be quick to address, though 🙂 @kennyadsl, do you want also to take another look?

app/models/solidus_stripe/gateway.rb
kennyadsl: What about changing the commit description to be clear that we are never using it in the current flow and why?

app/models/solidus_stripe/gateway.rb
```diff
@@ -34,11 +34,19 @@ def initialize(options)
     # @option options [Spree::Payment] :originator the payment on which the #capture is being performed (required)
     #
     # @return ActiveMerchant::Billing::Response
-    def capture(_amount_in_cents, _transaction_id, options = {})
```
kennyadsl, line 46: Please, can we update the YARD `@param _transaction_id [Object] (currently ignored)` option?

app/models/solidus_stripe/gateway.rb
```diff
@@ -95,14 +105,15 @@ def purchase(amount_in_cents, _source, options = {})
     # @param _options [Hash] (ignored)
     #
     # @return ActiveMerchant::Billing::Response
-    def void(_transaction_id, source, _options = {})
+    def void(payment_intent_id, _source, _options = {})
```
kennyadsl, line 93: same here for the YARD update

app/models/solidus_stripe/gateway.rb
```diff
@@ -126,12 +137,9 @@ def void(_transaction_id, source, _options = {})
     #   being performed (required if source is nil)
     #
     # @return ActiveMerchant::Billing::Response
-    def credit(amount_in_cents, _source, _transaction_id, options = {})
```
kennyadsl, line 111: same here for the YARD update

app/controllers/solidus_stripe/intents_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+    flash[:error] = t(
+      intent_status,
+      scope: 'solidus_stripe.payment_intent_status_flash_messages',
+      default: 'Something went wrong.',
+    )
+
+    redirect_to main_app.checkout_state_path(:payment)
+  end
+
+  def advance_order_with_notice(intent_status)
```
kennyadsl: this one seems to be unused.

app/controllers/solidus_stripe/intents_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  private
+
+  def ensure_state_is(object, state)
+    return if object.state.to_s == state.to_s
+
+    raise "unexpected object state #{object.state}, should have been #{state}"
+  end
+
+  def go_back_to_payment_with_error(intent_status)
```
kennyadsl: This also seems unused.

app/models/solidus_stripe/payment_method.rb
```diff
@@ -4,40 +4,153 @@ module SolidusStripe
...
-    alias cart_partial_name partial_name
-    alias product_page_partial_name partial_name
+      # @param payment [Spree::Payment] the payment we want to know if can be captured
+      # @return [Boolean] true when the payment is in the pending or checkout states
+      def can_capture?(payment)
+        payment.pending? || payment.checkout?
+      end
+
+      # @param payment [Spree::Payment] the payment we want to know if can be voided
+      # @return [Boolean] true when the payment is not failed, voided or invalid
```
kennyadsl: Is it intentional that `&& !payment.completed?` is not mentioned in the comment?

app/models/solidus_stripe/payment_method.rb
```diff
@@ -4,40 +4,153 @@ module SolidusStripe
...
+      def can_void?(payment)
+        payment.can_void? && !payment.completed?
+      end
 
-    def payment_source_class
-      PaymentSource
+      # Indicates whether its possible to credit the payment.  Note that most
```
kennyadsl: nit: double space after `payment.`

app/models/solidus_stripe/payment_method.rb
kennyadsl: Isn't `processing` another option?

lib/generators/solidus_stripe/install/templates/config/initializers/solidus_stripe.rb
kennyadsl: We discussed a solution in a pairing review session: using the `payment_intent_id` instead. I think Elia will update the PR soon with that.

kennyadsl (commented):
Left a round of comments after a partially joint review. @elia Let me know when I can review again.

app/models/solidus_stripe/payment_method.rb
waiting-for-dev: Yeah, it's the one used when the payment element has been rendered but not filled yet. Probably it won't be used at any point, but it makes sense to add it.

app/models/solidus_stripe/payment_method.rb
elia: Eventually removed the whole comment, which was copied verbatim from solidus and didn't add any value.

.circleci/config.yml
elia: Yes, we'll try to switch to VCR in the future, but we'll need to ensure not to leak credentials in the recorded cassettes, and contributors will also need credentials in order to record new cassettes. But it's definitively something we should look into 👍

app/models/solidus_stripe/gateway.rb
elia: Now I'm wishing for a yard/rdoc linter that can tell me the comment it's out of date 😒

app/models/solidus_stripe/gateway.rb
elia: It's not our decision, is how Solidus is implemented and received it from Spree/ActiveMerchant. I would also say that having it directly reduces the number of hops we need to do in order to get the value, which itself is a positive side effect.

lib/generators/solidus_stripe/install/templates/config/initializers/solidus_stripe.rb
elia: After a joint review with @kennyadsl we ended up using the payment intent id as the shared reference as it's seems sharing it has no security implications.

app/models/solidus_stripe/gateway.rb
elia: Added a commit that cleanup the yard documentation and adds it to the CI, failing if a warning is printed.

app/models/solidus_stripe/gateway.rb
elia: (see answer above)

app/models/solidus_stripe/gateway.rb
elia: (see above)

app/models/solidus_stripe/payment_method.rb
elia: Changed it to just be `pending?`, after all we only want to capture authorized payments

app/models/solidus_stripe/payment_method.rb
elia: Changed it to just be `pending?`, after all we only want to void authorized payments

app/models/solidus_stripe/payment_method.rb
elia: Here we want to retrieve any payment that is not in a final state, and those are the ones listed above. That said, it's entirely possible that we'll review all of this once we add profiles support. The whole class is mostly a placeholder and very unstable until we'll have the full picture (especially since we're not planning to release the extension without profiles support).

app/models/solidus_stripe/payment_method.rb
elia: Can you please expand on how it could impact security?

As explained above all of the code is still very up for change, and for now I'd like to keep the test coverage really light so it doesn't get in the way of the radical changes that are likely with the introduction of saved profiles and (possibly) webhook support.

Once things will be more stable we'll be able to look at the coverage and aim for higher percents.

Thoughts?

app/models/solidus_stripe/payment_method.rb
elia: With wallet support we'll have payments in checkout state for this payment method, but we can remove it for now I guess.

app/models/solidus_stripe/payment_method.rb
```diff
@@ -7,37 +7,146 @@ class PaymentMethod < ::Spree::PaymentMethod
...
+    concerning :Order do
+      def find_in_progress_payment_for(order)
+        payments = order.payments
+          .where.not(state: %w[completed invalid void]) # in_progress
+          .where(payment_method: self)
+          .order(:created_at)
+          .entries
+
+        if payments.size > 1
+          payments[1...-1].each(&:invalidate!)
```
loicginoux: If you mean to invalidate all potential existing payments and keep only the first one 
```suggestion
          payments[1..-1].each(&:invalidate!)
```

```ruby
[7] pry(#<SolidusStripe::PaymentMethod>)> a = [1,2,3]
[
    [0] 1,
    [1] 2,
    [2] 3
]
[8] pry(#<SolidusStripe::PaymentMethod>)> a[1..-1]
[
    [0] 2,
    [1] 3
]
[9] pry(#<SolidusStripe::PaymentMethod>)> a[1...-1]
[
    [0] 2
]
```

elia: Good catch! I actually meant to pick them all except the last one, but for some reason I used `1` instead of `0` as the first number, fixing!

app/models/solidus_stripe/payment_method.rb
```diff
@@ -7,37 +7,143 @@ class PaymentMethod < ::Spree::PaymentMethod
...
+    concerning :Order do
+      def find_in_progress_payment_for(order)
+        payments = order.payments
+          .where.not(state: %w[completed invalid void]) # in_progress
+          .where(payment_method: self)
+          .order(:created_at)
+          .entries
+
+        *old_payments, payment = payments
+        old_payments.each(&:invalidate!)
```
loicginoux, line 59: because you transition payment directly to processing at line 90, you might have an old payment in `processing` state that you will here try to invalidate. Processing payments cannot be invalidated so this line can fails. Maybe you could try both `void` (or cancel?) and `invalidate` with something like 
``` 
old_payments.each do |old_payment|
           old_payment.invalidate! if old_payment.can_invalidate?
           old_payment.void! if old_payment.can_void?
end
```

app/models/solidus_stripe/payment_method.rb
```diff
@@ -7,37 +7,143 @@ class PaymentMethod < ::Spree::PaymentMethod
...
         end
       end
 
-      self
+      def find_or_create_in_progress_payment_for(order)
+        payment = find_in_progress_payment_for(order)
+        intent = find_intent_for(payment) if payment
+
+        payment = nil if intent.nil?
+        payment = nil unless intent&.status == 'requires_payment_method'
```
loicginoux, line 99: From my tests here it seems like the intent you create during `create_in_progress_payment_for(order)` has always the status 'requires_source' (waiting to enter card details, I understand). So if you run multiple times `find_or_create_in_progress_payment_for` it will always recreate a payment + payment intent because you nulify the payment on this line. I think that's not what we want, we should void recreating intents whenever possible. What is the purpose of this line exactly ?

app/models/solidus_stripe/payment_method.rb
elia: Oh right!

The underlying issue is that to this date we don't have test coverage for this, but I don't want to commit to strict coverage until profiles are added to the mix, as things can still change considerably with that addition.

app/models/solidus_stripe/gateway.rb
waiting-for-dev: But we could ignore the argument, just as before (`_payment_intent_id`). It'd make the method easier to understand, as you wouldn't need to reach the conclusion we're discussing now. But that's a non-blocking from my side.

app/models/solidus_stripe/payment_method.rb
waiting-for-dev: Unless I'm missing something, that is a fix for https://github.com/solidusio/solidus_stripe/pull/158#issuecomment-1436840113, right? 🤔 

> I'd like to keep the test coverage really light so it doesn't get in the way of the radical changes that are likely with the introduction of saved profiles and (possibly) webhook support.

If that's the plan, I think we should be clear about that and create the appropriate tickets. Otherwise, it's too easy to forget and end up with a poorly tested extension.

app/models/solidus_stripe/payment_method.rb
waiting-for-dev: I see. That would be a good candidate to move upstream as a scope 👍 But let's move forward now.

app/models/solidus_stripe/payment_method.rb
waiting-for-dev: TBH, I don't feel comfortable with this approach. IMHO, the test suite should be something to change along the way. Otherwise, we're moving on top of shifting sands, and it's too easy to avoid covering some behavior on an ex-post read. At least, I think we should keep a list of things that need to be accounted for.

app/models/solidus_stripe/payment_method.rb
```diff
@@ -7,37 +7,143 @@ class PaymentMethod < ::Spree::PaymentMethod
...
-    def payment_profiles_supported?
-      true
+    concerning :Payment do
+      def create_profile(payment)
+        payment_intent = find_intent_for(payment)
+
+        if payment_intent && payment_intent.customer.blank?
+          payment.payment_method.gateway.request do
+            payment_intent.customer = Stripe::Customer.new email: payment.order.email
+            payment_intent.customer.save
```
loicginoux, line 113:
```
NOTE: Stripe::Customer#save is deprecated; use update instead. It will be removed on or after 2022-11-01.
```
https://github.com/stripe/stripe-ruby/blob/master/CHANGELOG.md#deprecated

app/models/solidus_stripe/payment_method.rb
```diff
@@ -7,37 +7,143 @@ class PaymentMethod < ::Spree::PaymentMethod
...
 
-    def payment_profiles_supported?
-      true
+    concerning :Payment do
+      def create_profile(payment)
+        payment_intent = find_intent_for(payment)
+
+        if payment_intent && payment_intent.customer.blank?
+          payment.payment_method.gateway.request do
+            payment_intent.customer = Stripe::Customer.new email: payment.order.email
```
loicginoux, line 112: From my tests, this does not work for several reasons:
- `Stripe::Customer.new` does not do anything.
- You cannot add a customer on an existing intent. You have to create the inetnt with the customer directly.
```bash
Stripe::InvalidRequestError: You cannot modify `customer` on a PaymentIntent once it already has been set. To fulfill a payment with a different Customer, cancel this PaymentIntent and create a new one.
from ~/.rvm/gems/ruby-2.7.7@epicery/gems/stripe-8.2.0/lib/stripe/stripe_client.rb:722:in `handle_error_response'
```
- You should not always create a new customer but find an existing customer if it exists. Something like:
```
            payment_intent.customer = find_or_create_customer(payment)
```
and then something like:
```ruby
      def find_or_create_customer(payment)
        res = payment.payment_method.gateway.request do
          Stripe::Customer.search({
            query: "email: '#{payment.order.email}'"
          }, stripe_version: '2020-08-27') # search is not supported for previous versions
        end
        if res[:data] && res[:data].first
          customer = res[:data].first
        else
          customer = payment.payment_method.gateway.request do
            Stripe::Customer.create({ email: payment.order.email })
          end
        end
        customer
      end
```

loicginoux (changes requested):

app/models/solidus_stripe/payment_method.rb
loicginoux: Maybe atatching a customer to the intent should be done in another ticket/PR. It is not mandatory for this PR I'd say...

README.md
elia: Much better!

README.md
elia: I'm not sure who this can be done through a background job, also I feel pretty confident that this should be ok having done that in multiple stores for one-page checkouts or to fast-forward steps that were unnecessary.

app/models/solidus_stripe/payment_method.rb
elia: @loicginoux extracted the comment for a later fix in https://github.com/solidusio/solidus_stripe/issues/200 so we don't forget.

The discussion about how to approach development was taken to a series of online meetings and beyond the scope of this comment thread. I'll leave an update at the end of this PR later.

app/models/solidus_stripe/payment_method.rb
elia: Tracked in #201 thanks!

app/models/solidus_stripe/payment_method.rb
elia: Added a todo item for this in #172 which already has some code for it

app/models/solidus_stripe/payment_method.rb
elia: also tracked in #172

elia (commented):
Added some answers and mostly tracked pending comments in new issues or todo items on #172. 

I'm going to merge this so we can stop rebasing on top of it and fix master having the payment intent id connected to the payment source as per internal team agreement.

Issue #179, opened 2023-02-22
Handle `payment_intent.succeeded` webhook event


waiting-for-dev:
We'll get a `payment_intent.succeeded` webhook event whenever:

- A payment intent is captured from Stripe's dashboard
- A Solidus payment is manually captured from the admin interface (see #158)
- TODO: auto-capture is on, and an order is completed through a regular checkout (see #178)

We must update the Solidus counterpart for the first scenario by triggering the `complete` transition from the Solidus payments state machine. However, we need to be resilient because of the other two scenarios and the lack of guarantee that a webhook will only be received once and do nothing if the payment is already completed.

There's one caveat: Stripe allows to partially capture a payment, while Solidus doesn't. For that case, we'll update the `Spree::Payment#amount` field, and let's consider if we can add a log entry.

Partially replaces #160

waiting-for-dev:
Leaving out handling partial captures, as that might require more thought. See https://github.com/solidusio/solidus_stripe/issues/221.

PR #195, opened 2023-02-27
Make webhook testing more realistic and refactor its setup

waiting-for-dev:
## Summary

We're splitting the webhook testing helpers into three parts:

- `EventWithContextFactory` can create a stripe event from different sources (hash, Stripe object) and in different representations (JSON, `Stripe::Event` instance, header value in a request).

- `HashFixtures` hardcodes a Hash representation of a Stripe event. We can use it with `EventWithContextFactory.from_hash` to consume it.

- `RequestHelper` provides helper methods for request specs to simulate webhook requests from an instance of `EventWithContextFactory`.

We leverage creating a Stripe webhook event from a Stripe object on request specs. The creation of the Stripe object will call Stripe's API, giving us more assurance that we're working with actual data; plus, it allows us to create events from objects created from within the library code (i.e., payment intents). Although the generated scenario won't be 100% real, as we still need to wrap the object with boilerplate data, it's a good compromise against using [`stripe-cli`](https://stripe.com/docs/webhooks/test) on the test suite, which would require a complex setup mainly on CI.

Conversely, we leverage the Hash fixture for unit tests on `SolidusStripe::Webhook::Event` to avoid performing requests there.

This PR will pave the way for #179 and all the other webhook events we're going to implement.

## Checklist

### DIFF
--- a/spec/lib/solidus_stripe/webhook/event_spec.rb
+++ b/spec/lib/solidus_stripe/webhook/event_spec.rb
@@ -24,13 +24,16 @@ RSpec.describe SolidusStripe::Webhook::Event do
   end
 
   describe ".from_request" do
+    let(:context) do
+      SolidusStripe::Webhook::EventWithContextFactory.from_data(
+        data: SolidusStripe::Webhook::DataFixtures.charge_succeeded
+      )
+    end
+
     context "with a valid event" do
       it "returns an event" do
-        payload = JSON.generate(SolidusStripe::WebhookFixtures.charge_succeeded)
-        fixture = SolidusStripe::WebhookFixtures.new(payload: payload)
-
         event = described_class.from_request(
-          payload: payload, signature_header: fixture.signature_header, secret: fixture.secret
+          payload: context.json, signature_header: context.signature_header, secret: context.secret
         )
 
         expect(event).to be_a(described_class)
@@ -39,12 +42,20 @@ RSpec.describe SolidusStripe::Webhook::Event do
 
     context "when the signature is invalid" do
       it "returns nil" do
-        payload = JSON.generate(SolidusStripe::WebhookFixtures.charge_succeeded)
-        fixture = SolidusStripe::WebhookFixtures.new(payload: payload)
         signature_header = "t=1,v1=1"
 
         event = described_class.from_request(
-          payload: payload, signature_header: signature_header, secret: fixture.secret
+          payload: context.json, signature_header: signature_header, secret: context.secret
+        )
+
+        expect(event).to be(nil)
+      end
+    end
+
+    context "when the tolerance has expired" do
+      it "returns nil" do
+        event = described_class.from_request(
+          payload: context.json, signature_header: context.signature_header, secret: context.secret, tolerance: 0
         )
 
         expect(event).to be(nil)
@@ -53,11 +64,8 @@ RSpec.describe SolidusStripe::Webhook::Event do
 
     context "when the payload is malformed" do
       it "returns nil" do
-        payload = "invalid"
-        fixture = SolidusStripe::WebhookFixtures.new(payload: payload)
-
         event = described_class.from_request(
-          payload: payload, signature_header: fixture.signature_header, secret: fixture.secret
+          payload: "invalid", signature_header: context.signature_header, secret: context.secret
         )
 
         expect(event).to be(nil)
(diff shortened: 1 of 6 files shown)


waiting-for-dev:
CI is failing because it doesn't have access to the stripe context introduced in https://github.com/solidusio/solidus_stripe/pull/158

waiting-for-dev:
> CI is failing because it doesn't have access to the stripe context introduced in #158

Chaning base branch for now

elia:
@waiting-for-dev you should be able to target master once https://github.com/solidusio/solidus_stripe/pull/196 is merged

spec/lib/solidus_stripe/webhook/event_spec.rb
```diff
@@ -66,40 +64,40 @@
...
     end
   end
 
   describe "#payload" do
-    let(:charge_suceeded_event) do
-      Stripe::Event.construct_from(
-        SolidusStripe::WebhookFixtures.charge_succeeded
+    let(:context) do
+      SolidusStripe::Webhook::EventWithContextFactory.new(
+        hash: SolidusStripe::Webhook::HashFixtures.charge_succeeded
```
elia: Thoughts on moving back to more explicit names by embedding the call to `.object` in the `let`?

```rb
let(:charge_suceeded_event) do
  SolidusStripe::Webhook::EventWithContextFactory.new(
    hash: SolidusStripe::WebhookFixtures.charge_succeeded
  ).object
end

File: spec/requests/solidus_stripe/webhooks_controller_spec.rb [hunk 2]
```diff
@@ -1,51 +1,48 @@
...
         expect(response).to have_http_status(:ok)
       end
     end
 
     context "when the event can't be generated" do
-      let(:payload) { "invalid" }
-
       it "returns a 400 status code" do
-        post "/solidus_stripe/webhooks", params: payload, headers: { signature_header_key => fixture.signature_header }
+        webhook_request(context, timestamp: Time.zone.yesterday.midnight)
```
elia: I'm confused about why this will return a 400, is it because the timestamp is in the past?

waiting-for-dev:
> @waiting-for-dev you should be able to target master once #196 is merged

Thanks, @elia. It's now 🟢  and ready for review (cc @kennyadsl)

kennyadsl (approved):
Not blocking, but I'd add some documentation on top of helpers' class definition to describe how to use them. 👍

waiting-for-dev:
> Not blocking, but I'd add some documentation on top of helpers' class definition to describe how to use them. +1

Done. Would you mind taking another look? :pray:

spec/support/solidus_stripe/webhook/event_with_context_factory.rb
```diff
@@ -0,0 +1,71 @@
...
+          "request" => {
+            "id" => "req_3MRUo1JvEPu9yc7w0",
+            "idempotency_key" => "idempotency_key"
+          },
+          "type" => type
+        }
+      end
+      private_class_method :hash_base
+
+      attr_reader :hash, :timestamp, :secret
```
elia: nit: Not that instances of this class will ever be used as hash-keys, but `#hash` is a special method name which use might be puzzling at first sight, thoughts on renaming it to `data`, `payload`, or something along those lines?

spec/support/solidus_stripe/webhook/request_helper.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Webhook
```
elia: Should this be namespaces in `SolidusStripe` as well? or we have reasons not to do that?

kennyadsl (approved):

elia (changes requested):
Great to see this getting more structure! 🙌 

A few questions, as usual those marked as _nits_ are not blockers for merging

spec/support/solidus_stripe/webhook/event_with_context_factory.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      def initialize(hash:, timestamp: Time.zone.now, secret: "whsec_123")
+        @hash = hash
+        @timestamp = timestamp
+        @secret = secret
+      end
+
+      def object
```
loicginoux: Similarly to comment about `hash` above, it could be more explicit to call this method and instance variable `event` or `stripe_event` rather than  `object` to make things more clear and specific

spec/lib/solidus_stripe/webhook/event_spec.rb
waiting-for-dev: Yeah, that makes sense, as we're not using any other property of the context object. Thanks for the suggestion. Done!

spec/requests/solidus_stripe/webhooks_controller_spec.rb
waiting-for-dev: Exactly. Stripe events are discarded if they're more than 5 minutes old by default. That's used to [mitigate replay attacks](https://stripe.com/docs/webhooks/signatures#replay-attacks).

spec/support/solidus_stripe/webhook/event_with_context_factory.rb
waiting-for-dev: That's an excellent point, and I overlooked that. I renamed it to `#data`, as `#payload` is expected to be the JSON string from the Stripe API (which could be confusing).

spec/support/solidus_stripe/webhook/request_helper.rb
waiting-for-dev: No reason. Thanks for catching that! Fixed.

spec/support/solidus_stripe/webhook/event_with_context_factory.rb
waiting-for-dev: Thanks for your suggestion, @loicginoux. The thing is that I want to encapsulate different representations of the Stripe event in different formats. So all of `json`, `hash` (now `data`), and `object` are stripe events. I used `object` as that's the Stripe naming for [its resources](https://www.rubydoc.info/gems/stripe/Stripe/StripeObject).

Maybe I could call it `instance`, but probably that would be very similar. Another option would be a verbose `stripe_object`. Thoughts?

waiting-for-dev: I'd ok with `stripe_object`, as that would probably remove the confusion you found. Waiting for your feedback.

spec/requests/solidus_stripe/webhooks_controller_spec.rb
elia: Thoughts on explaining that in the example description? (not blocking)

elia (approved):
Left one more (completely optional) suggestion 👍👍👍

spec/support/solidus_stripe/webhook/event_with_context_factory.rb
loicginoux: mhhh.. I dind't catch that, so yes, IMO I would say stripe_object is better but it's your choice.

waiting-for-dev: Thanks, @loicginoux. Done!

spec/requests/solidus_stripe/webhooks_controller_spec.rb
waiting-for-dev: @elia, in the request spec I'm intentionally vague (`when the event can't be generated`), because I'm converting the different scenarios on the `StripeEvent::Webhook::Event.from_request` [specs](https://github.com/solidusio/solidus_stripe/blob/9ea732b97be26683b192545b9e7300289192ea37/spec/lib/solidus_stripe/webhook/event_spec.rb#L26) (which are more performant).

waiting-for-dev: But thanks for pointing out. In fact, I just realized that we were missing testing expiration at the unit level, so I've just added it :slightly_smiling_face:

PR #202, opened 2023-03-02
Consolidate how Rails version is fetched for dummy-app and sandbox

waiting-for-dev:
## Summary

We use now the same that it's used for the sandbox.

## Checklist  

### DIFF
--- a/bin/dummy-app
+++ b/bin/dummy-app
@@ -14,7 +14,8 @@ function unbundled {
 test "$DB" = "sqlite" && export DB="sqlite3"
 
 rm -rf ./dummy-app
-rails ${RAILS_VERSION:+_${RAILS_VERSION}_} new dummy-app \
+rails_version=`bundle exec ruby -e'require "rails"; puts Rails.version'`
+rails _${rails_version}_ new dummy-app \
   --database=${DB:-sqlite3} \
   --skip-git \
   --skip-rc


bin/dummy-app
```diff
@@ -14,7 +14,8 @@ function unbundled {
 test "$DB" = "sqlite" && export DB="sqlite3"
 
 rm -rf ./dummy-app
-rails ${RAILS_VERSION:+_${RAILS_VERSION}_} new dummy-app \
+rails_version=`bundle exec ruby -e'require "rails"; puts Rails.version'`
+rails _${rails_version}_ new dummy-app \
```
elia, line 18: I've had issues with using this setup calling `bundle exec`, inside the script, thoughts on doing it the other way around and updating the sandbox instead?

Using the env var seems more adaptable to different situations and the Gemfile would be based on an environment variable anyway.

elia (changes requested):

bin/dummy-app
waiting-for-dev: The issue is that in the script context, `$RAILS_VERSION` is expected to be an actual version, like `7.0` or `6.2`. However, in other places of the Solidus ecosystem is expected to be a semver reference, like `~> 7.0`. See [docker config on main Solidus](https://github.com/solidusio/solidus/blob/c09ead361b6b705897bf6a05aa38cef24d7d3c0d/docker-compose.yml#L36) or the [CircleCI orb](https://github.com/solidusio/circleci-orbs-extensions/blob/a86ae2bdf4b0b99466e5b36c96a5edcd3ad43b8d/src/commands/test-branch.yml#L28) as examples. (I use docker for development, and my docker-compose file had the env var; that was how it hit me :sweat_smile:).

Do you recall which problem you found with `bundle exec`? At first sight, it looks like the safest way to get the actual version that we're using.

elia: No I think it was related to versions of bundler one nested into the other, but I don't have any reference, let's go with this code and if the problem comes up again we can switch to another solution 👍

elia (approved):
🙌

bin/dummy-app
waiting-for-dev: All right, thanks, @elia!

PR #213, opened 2023-03-09
Make webhook endpoint specific to the payment method instance

### LINKED ISSUE #161: Make webhook handling connected to the payment method
## What

The secret for verifying the webhooks is specific to the stripe account and should be retrieved from the payment method preferences.

## Why

Although it's unlikely to have multiple Stripe accounts it's correct to store an account specific secret as a `Spree::PaymentMethod` preference along with the other credentials.

waiting-for-dev:
## Summary

The work on this PR allows registering different webhook endpoints for each instance of Solidus Stripe payment methods. That's a scenario that can help support marketplaces.

We're automatically generating a slug when a Stripe payment method is created. Users must use that in the default `/solidus_stripe/webhooks/:slug` endpoint to get the specific URL for a payment method instance. The slug is stored on a separate `SolidusStripe::WebhookEndpoint` model with a one-to-one relationship with `SolidusStripe::PaymentMethod`. There's no single entry point to create payment methods, so we must use an ActiveRecord callback to persist the slug. We're using a slug and not the payment method id to be resilient against database changes and to avoid guessing about valid endpoint URLs.

As a consequence of bringing the webhook endpoint URL to the payment method instance, we're also moving the configuration of the webhook signing secret from a generic Rails credential to a payment method preference.

We're now providing the slug to the `SolidusStripe::Event.from_request` method to retrieve the associated payment method and its signing secret. The payment method is now bundled as part of the created Omnes event so that subscribers can act differently in every case. For the test setup, the payment method is now part of the `EventWithContextFactory` instances.

Closes #161

## Checklist

### DIFF
--- a/app/controllers/solidus_stripe/webhooks_controller.rb
+++ b/app/controllers/solidus_stripe/webhooks_controller.rb
@@ -12,7 +12,8 @@ module SolidusStripe
     respond_to :json
 
     def create
-      event = Webhook::Event.from_request(payload: request.body.read, signature_header: signature_header)
+      event = Webhook::Event.from_request(payload: request.body.read, signature_header: signature_header,
+        slug: params[:slug])
       return head(:bad_request) unless event
 
       Spree::Bus.publish(event) && head(:ok)
--- a/app/models/solidus_stripe/payment_method.rb
+++ b/app/models/solidus_stripe/payment_method.rb
@@ -8,9 +8,27 @@ module SolidusStripe
     preference :stripe_intents_flow, :string, default: 'setup'
     preference :skip_confirmation_for_payment_intent, :boolean, default: true
 
+    # @attribute [rw] preferred_webhook_endpoint_signing_secret The webhook endpoint signing secret
+    #  for this payment method.
+    # @see https://stripe.com/docs/webhooks/signatures
+    preference :webhook_endpoint_signing_secret, :string
+
     validates :available_to_admin, inclusion: { in: [false] }
     validates :preferred_setup_future_usage, inclusion: { in: ['', 'on_session', 'off_session'] }
     validates :preferred_stripe_intents_flow, inclusion: { in: ['payment', 'setup'] }
+
+    has_one :webhook_endpoint,
+      class_name: 'SolidusStripe::WebhookEndpoint',
+      inverse_of: :payment_method,
+      dependent: :destroy
+
+    after_create :assign_webhook_endpoint
+
+    # @!attribute [r] webhook_endpoint_slug
+    #   @return [String] The slug of the webhook endpoint for this payment method.
+    delegate :slug,
+      to: :webhook_endpoint,
+      prefix: true
 
     concerning :Configuration do
       def partial_name
@@ -112,6 +130,14 @@ module SolidusStripe
       when /^seti_/
         "https://dashboard.stripe.com#{path_prefix}/setup_intents/#{intent_id}"
       end
+    end
+
+    private
+
+    def assign_webhook_endpoint
+      create_webhook_endpoint!(
+        slug: WebhookEndpoint.generate_slug
+      )
     end
   end
 end
--- a/app/models/solidus_stripe/webhook_endpoint.rb
+++ b/app/models/solidus_stripe/webhook_endpoint.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+module SolidusStripe
+  # Represents a webhook endpoint for a {SolidusStripe::PaymentMethod}.
+  #
+  # A Stripe webhook endpoint is a URL that Stripe will send events to. A store
+  # could have multiple Stripe payment methods (e.g., a marketplace), so we need
+  # to differentiate which one a webhook request targets.
+  #
+  # This model associates a slug with a payment method. The slug is appended
+  # to the endpoint URL (`.../webhooks/:slug`) so that we can fetch the
+  # correct payment method from the database and bind it to the generated
+  # `Spree::Bus` event.
+  #
+  # We use a slug instead of the payment method ID to be resilient to
+  # database changes and to avoid guessing about valid endpoint URLs.
+  class WebhookEndpoint < ::Spree::Base
+    belongs_to :payment_method,
+      class_name: 'SolidusStripe::PaymentMethod'
+
+    # @api private
+    def self.generate_slug
+      SecureRandom.hex(16).then do |slug|
+        exists?(slug: slug) ? generate_slug : slug
+      end
+    end
+
+    # Finds the payment method associated with the given slug.
+    #
+    # @param slug [String]
+    # @raise [ActiveRecord::RecordNotFound] if no payment method is found
+    # @return [SolidusStripe::PaymentMethod]
+    def self.payment_method(slug)
+      find_by!(slug: slug).payment_method
+    end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,5 +5,5 @@ SolidusStripe::Engine.routes.draw do
     get :payment_confirmation, controller: :intents
     get :setup_confirmation, controller: :intents
   end
-  resources :webhooks, only: :create, format: false
+  post '/webhooks/:slug', format: false, to: 'webhooks#create'
 end
--- a/README.md
+++ b/README.md
@@ -47,29 +47,33 @@ bin/rails generate solidus_stripe:install
 
 This library makes use of some [Stripe webhooks](https://stripe.com/docs/webhooks).
 
-On development, you can [test them by using Stripe CLI](https://stripe.com/docs/webhooks/test).
+Every Solidus Stripe payment method you create will get a slug assigned. You
+need to append it to a generic webhook endpoint to get the URL for that payment
+method. For example:
 
-Before going to production, you'll need to [register the
-`/solidus_stripe/webhooks` endpoint with
+```ruby
+SolidusStripe::PaymentMethod.last.webhook_endpoint_secret
+# "365a8435cd11300e87de864c149516e0"
+```
+
+For the above example, and if you mounted the `SolidusStripe::Engine` routes on
+the default scope, the webhook endpoint would look like:
+
+```
+/solidus_stripe/webhooks/365a8435cd11300e87de864c149516e0
+```
+
+Besides, you also need to configure the webhook signing secret for that payment
+method. You can do that through the `webhook_endpoint_signing_secret`
+preference on the payment method.
+
+Before going to production, you'll need to [register the webhook endpoint with
 Stripe](https://stripe.com/docs/webhooks/go-live), and make sure to subscribe
-to the following events:
+to the events listed in [the `SolidusStripe::Webhook::Event::CORE`
+constant](https://github.com/solidusio/solidus_stripe/blob/master/lib/solidus_stripe/webhook/event.rb).
 
-[TBD]
-
-In both environments, you'll need to create a
-`solidus_stripe.webhook_endpoint_secret` credential with [the webhook signing
-secret](https://stripe.com/docs/webhooks/signatures):
-
-```bash
-# For development, add `--environment development`
-bin/rails credentials:edit
-```
-
-```yaml
-# config/credentials.yml.enc
-solidus_stripe:
-  webhook_endpoint_secret: "whsec_..."
-```
+On development, you can
+[test webhooks by using Stripe CLI](https://stripe.com/docs/webhooks/test).
 
 ## Usage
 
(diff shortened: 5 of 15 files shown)


app/models/solidus_stripe/webhook_endpoint.rb
```diff
@@ -0,0 +1,37 @@
...
+  # We use a suffix instead of the payment method ID to be resilient to
+  # database changes and to avoid guessing about valid endpoint URLs.
+  class WebhookEndpoint < ::Spree::Base
+    belongs_to :payment_method,
+      class_name: 'SolidusStripe::PaymentMethod'
+
+    # @api private
+    def self.generate_suffix
+      SecureRandom.hex(16).then do |suffix|
+        exists?(suffix: suffix) ? generate_suffix : suffix
```
kennyadsl: What about also enforcing this with a validation? Also, we could consider adding a unique index on the database table, both for performances and to enforce this validation on the db level. Thoughts?

kennyadsl (commented):
Nice change, Marc. I'm not sure to get the reason behind using the `suffix` naming, though to me, it seems more like a slug.

app/models/solidus_stripe/payment_method.rb
```diff
@@ -8,10 +8,28 @@ class PaymentMethod < ::Spree::PaymentMethod
...
+    preference :webhook_endpoint_signing_secret, :string
+
     validates :available_to_admin, inclusion: { in: [false] }
     validates :preferred_setup_future_usage, inclusion: { in: ['', 'on_session', 'off_session'] }
     validates :preferred_stripe_intents_flow, inclusion: { in: ['payment', 'setup'] }
 
+    has_one :webhook_endpoint,
+      class_name: 'SolidusStripe::WebhookEndpoint',
+      inverse_of: :payment_method,
+      dependent: :destroy
```
elia, line 23: I'm not sure I understand why this is not a preference (e.g. `preference :webhook_suffix`), can you expand on that?

elia: Oh, I think it's for finding it quickly from the DB instead of loading up all the payment methods just to get the preference value, because it's stored in a YAML, is that correct?

If that's the case I think it would be ok to expect a store only to have a handful of payment methods and shouldn't have any impact on performance. Thoughts?

app/controllers/solidus_stripe/webhooks_controller.rb
```diff
@@ -12,7 +12,8 @@ class WebhooksController < Spree::BaseController
     respond_to :json
 
     def create
-      event = Webhook::Event.from_request(payload: request.body.read, signature_header: signature_header)
+      event = Webhook::Event.from_request(payload: request.body.read, signature_header: signature_header,
```
elia, line 15: Thoughts on refactoring this? 
Maybe loading the payment method first (using params[:suffix]) and then passing it to the method along with the request body.

config/routes.rb
```diff
@@ -5,5 +5,5 @@
     get :payment_confirmation, controller: :intents
     get :setup_confirmation, controller: :intents
   end
-  resources :webhooks, only: :create, format: false
+  post '/webhooks/:suffix', format: false, to: 'webhooks#create'
```
elia: I was wondering if it made sense to scope everything within a sort if payment method id/code that identifies it without being the sequential number, thoughts on this?

elia (commented):
Left some comments and ideas, not necessarily things to follow upon or to tackle in this PR, but was interested in your take

waiting-for-dev:
> Nice change, Marc. I'm not sure to get the reason behind using the `suffix` naming, though to me, it seems more like a slug.

@kennyadsl, I think you're right. I renamed it to `slug`. Thanks for pointing that out.

app/models/solidus_stripe/webhook_endpoint.rb
waiting-for-dev: > What about also enforcing this with a validation? 

I think there's no need to add validation for that, as that's not part of user input. The library code generates that, so testing is the right place to "validate" it. What do you think?

> Also, we could consider adding a unique index on the database table, both for performances and to enforce this validation on the db level. Thoughts?

I agree. I added the index and made it unique. Probably we won't appreciate any performance difference (very few records for `Spree::PaymentMethod), and, in fact, making the index unique will impact negatively, but I like to have consistency at the DB layer unless there's a reason to skip. Thanks for your suggestion.

app/models/solidus_stripe/payment_method.rb
waiting-for-dev: @elia, thanks for that suggestion. Yeah, I considered having it as a preference, but besides what you pointed out, there's no way to ensure uniqueness, and it requires user action for something that should be automatic. What do you think?

app/controllers/solidus_stripe/webhooks_controller.rb
waiting-for-dev: IMHO, that's in the correct layer. The `.from_request` method gets all the request information (the request body, the signature header, and the slug) and transforms it into a webhook event instance. For instance, if it can't find a payment method with that slug, it will return `nil` just like when the payload is malformed, or the signature header is invalid. The controller leverages that to return a bad response code to Stripe. What do you think?

config/routes.rb
waiting-for-dev: Not sure I get the point. Do you mind using the `suffix` also to scope the `get :payment_confirmation` & `get :setup_confirmation` routes?

app/models/solidus_stripe/payment_method.rb
```diff
@@ -170,5 +188,13 @@ def stripe_dashboard_url(payment)
...
       "https://dashboard.stripe.com#{path_prefix}/payments/#{intent_id}"
     end
+
+    private
+
+    def assign_webhook_endpoint
+      create_webhook_endpoint!(
+        slug: WebhookEndpoint.generate_slug
+      )
+    end
```
rainerdema, line 141: Considering the low probability of encountering duplicate slugs and the recent addition of an index to the slug column, it would be worth reassessing the current slug generation logic (by simply returning a random slug) and adding a rescue and retry strategy to handle any unique constraint violations? What do you think?

rainerdema (commented):
Great job, Marc! 👏  I left one small consideration, but overall, the changes look good!

app/models/solidus_stripe/payment_method.rb
elia: I'm conflicted, I wish prefs were more queryable, because uniqueness can be simply ensured through a validation, and probably generating a default value is equally easy (e.g. before_validate, before_create, a dynamic pref default etc.).

But having them as YAML make it all a bit clunky. This is basically an associated model just to have an extra column on payment method.

Final decision is yours, and as everything else we can return to it if we want to change, whatever you pick 👍

app/models/solidus_stripe/payment_method.rb
elia: Give it's after create and we have the payment method id we can even use that as the starting point and avoid the retry strategy altogether:

```rb
create_webhook_endpoint!(slug: Digest::MD5.hexdigest(id))
```

or use uuid which has [pretty strong guarantees against collision](https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions) which could or could not be less the probability of [cosmic-ray induced errors](https://stackoverflow.com/questions/2580933/cosmic-rays-what-is-the-probability-they-will-affect-a-program) (this SO post is fun 😅):

```rb
create_webhook_endpoint!(slug: SecureRandom.uuid)
```

waiting-for-dev: That's a great point. I guess you have in mind getting rid of the `.exists?` extra query, right? In a typical application, I'd agree. Still, here the migration gets copied into the host store, and they could decide to eliminate the unique constraint at the database layer. Creating the payment method will likely only happen once in most cases, so I think I prefer the solution that would work in any situation. What do you think?

app/controllers/solidus_stripe/webhooks_controller.rb
elia: Oh I think I came from the perspective of having `.from_request(…)` expecting a request as the main param, and maybe some options instead of all the parts relative to the request, and then something very railsy in the controller, like:

```rb
event = @payment_method.build_webhook_event(payload: request.body.read, signature_header: signature_header)

**Review by kennyadsl — APPROVED** [review]
Nice! Not blocking but there are some references of suffix in the commit and PR descriptions which might be good to update.

File: app/models/solidus_stripe/payment_method.rb (hunk 5 above)
Comment (rainerdema) [reply app/models/solidus_stripe/payment_method.rb:141]: >That's a great point. I guess you have in mind getting rid of the .exists? extra query, right?

Yes exactly 👍

>In a typical application, I'd agree. Still, here the migration gets copied into the host store, and they could decide to eliminate the unique constraint at the database layer. 

In my opinion, it's the developer's responsibility to ensure that any changes made to the migration, like the example mentioned, are also compatible with the rest of the code.

>Creating the payment method will likely only happen once in most cases, so I think I prefer the solution that would work in any situation. What do you think?

However, I do appreciate your perspective, and given the infrequent use of this function in the payment method creation process, I agree that it's reasonable to maintain this option 👍

**Review by rainerdema — APPROVED** [review]
💯

**Comment by waiting-for-dev** [comment]
> Nice! Not blocking but there are some references of suffix in the commit and PR descriptions which might be good to update.

Oh, thanks for pointing that out. Fixed!

**Review by elia — COMMENTED** [review]
I'm so sorry, I had this comments sitting in pending since Friday 😓 
Please ping me if you wanna discuss anything or another review and speed up the merging 🙏 thanks for your patience!

File: app/models/solidus_stripe/payment_method.rb (hunk 2 above)
Comment (waiting-for-dev) [reply app/models/solidus_stripe/payment_method.rb:23]: > can be simply ensured through a validation

Is that true? I mean, you can create a valid payment method record and later change the preferences via env variables. I think Solidus wouldn't validate that on app initialization, but as you say, it'd be clumsy and cumbersome.

File: app/controllers/solidus_stripe/webhooks_controller.rb (hunk 3 above)
Comment (waiting-for-dev) [reply app/controllers/solidus_stripe/webhooks_controller.rb:15]: I'd prefer not to have yet another method on the payment method model :sweat_smile: (although it's getting thinner), and I'd see it as going way beyond its responsibilities. However, as you suggested, we could pass the request parameter to the `Webhook::Event` class method. That would make the signature simpler, but I think that responsibilities are better defined as they are now:
- The Rails controller knows how to extract things from its `request` method.
- The `Webhook::Event.from_request` would be something more like `Webhook::Event.from_request_information`, less coupled to the Rails internals.
What do you think?

File: config/routes.rb (hunk 4 above)
Comment (elia) [reply config/routes.rb]: Yes I meant that, once we have a code identifying the payment method is more consistent to scope all the routes in the same way

File: app/models/solidus_stripe/payment_method.rb (hunk 2 above)
Comment (elia) [reply app/models/solidus_stripe/payment_method.rb:23]: In the end this is probably more of an upstream problem, in which prefs are a weird and not fully consistent tool. I took note and you're right, this is a good solution given the context 👍

File: config/routes.rb (hunk 4 above)
Comment (waiting-for-dev) [reply config/routes.rb]: I see your point. However, on the other side, it also makes sense to stick to the Rails default way unless there's a compelling reason, and that reason only exists for the webhook endpoint. Regardless, are you ok if we discuss that on a separate ticket? Ultimately, that's outside of the scope of this one's purpose.

File: app/controllers/solidus_stripe/webhooks_controller.rb (hunk 3 above)
Comment (elia) [reply app/controllers/solidus_stripe/webhooks_controller.rb:15]: That's ok, no need to change the method name, I guess what bothers me (but only a little) is passing the slug, thoughts on letting the controller fetch the payment method from it?

Also probably we can shorten the `signature_header` param to just `signature`:

```rb
Webhook::Event.from_request(
  payload: request.body.read, 
  signature: request.headers[SIGNATURE_HEADER], 
  payment_method: SolidusStripe::WebhookEndpoint.find_by!(slug: params[:slug]).payment_method
)
```

_Anyway consider all of the above as non-blocking nits_ 😊

config/routes.rb
elia: Sure thing, it's not blocking 👍

elia (approved):
Left another proposal for the webhook building in the controller, but feel free to discard it or keep it for another iteration (sorry again for blocking this with that pending review!).

config/routes.rb
waiting-for-dev: Cool! I created https://github.com/solidusio/solidus_stripe/issues/220. Thanks, @elia!

app/controllers/solidus_stripe/webhooks_controller.rb
waiting-for-dev: That would only work for some scenarios. If `find_by!` raises, we'd get a 500 response code. Now the query is done from the `.from_request` method, and it rescues the `ActiveRecord::NotFound` error alongside other decoding errors to return a 400 to Stripe.

I'll merge it as it is, but I'm more than happy to keep discussing it!

waiting-for-dev:
> Left another proposal for the webhook building in the controller, but feel free to discard it or keep it for another iteration (sorry again for blocking this with that pending review!).

No worries, @elia. I really appreciate your detailed review :slightly_smiling_face:

Issue #210, opened 2023-03-07
Extract a strategy class(es) from the parallel code implementing payment & setup intents


elia:
## Why

Moving the current prefs to a "strategy" class implementing the behavior opens up more customization possibilities for end users, and makes deviating from the defaults more intentional.

## What 

We have multiple code sites having parallel code for payment and setup intents, those are great starting points for molding the API of the strategy class. Eventually also the preferences regarding the flow and skipping the confirm step need to be moved to the strategy class.

### Related

#173 can also be solved by using the strategy class.

elia:
Superseded by #249

PR #214, opened 2023-03-10
Refactor intent related methods out of PaymentMethod

elia:
## Summary

We already had _concerns_ grouping methods in the PaymentMethod class, now we took the time to actually move them to their legitimate place in the intent classes.

This is to be considered a preliminary step toward encapsulating the strategy into a class.

Ref #210 

## Checklist

### DIFF
--- a/app/models/solidus_stripe/payment_method.rb
+++ b/app/models/solidus_stripe/payment_method.rb
@@ -163,13 +163,18 @@ module SolidusStripe
       false
     end
 
-    def stripe_dashboard_url(payment)
-      path_prefix = '/test' if preferred_test_mode
+    # Fetches the payment intent when available, falls back on the setup intent associated to the order.
+    def self.intent_id_for_payment(payment)
+      return unless payment
 
-      intent_id =
-        payment.transaction_id ||
+      payment.transaction_id || (
         SolidusStripe::PaymentIntent.where(order: payment.order).pick(:stripe_payment_intent_id) ||
         SolidusStripe::SetupIntent.where(order: payment.order).pick(:stripe_setup_intent_id)
+      )
+    end
+
+    def stripe_dashboard_url(intent_id)
+      path_prefix = '/test' if preferred_test_mode
 
       case intent_id
       when /^pi_/
--- a/app/models/solidus_stripe/payment_method.rb
+++ b/app/models/solidus_stripe/payment_method.rb
@@ -91,13 +91,16 @@ module SolidusStripe
     end
 
     # Fetches the payment intent when available, falls back on the setup intent associated to the order.
+    # @api private
+    # TODO: re-evaluate the need for this and think of ways to always go throught the intent classes.
     def self.intent_id_for_payment(payment)
       return unless payment
 
-      payment.transaction_id || (
-        SolidusStripe::PaymentIntent.where(order: payment.order, payment_method: payment.payment_method)&.pick(:stripe_payment_intent_id) ||
-        SolidusStripe::SetupIntent.where(order: payment.order, payment_method: payment.payment_method)&.pick(:stripe_setup_intent_id)
-      )
+      payment.transaction_id || SolidusStripe::PaymentIntent.where(
+        order: payment.order, payment_method: payment.payment_method
+      )&.pick(:stripe_intent_id) || SolidusStripe::SetupIntent.where(
+        order: payment.order, payment_method: payment.payment_method
+      )&.pick(:stripe_intent_id)
     end
 
     def stripe_dashboard_url(intent_id)
--- a/app/models/solidus_stripe/payment_method.rb
+++ b/app/models/solidus_stripe/payment_method.rb
@@ -82,77 +82,6 @@ module SolidusStripe
       end
     end
 
-    concerning :SetupIntents do
-      def find_or_create_setup_intent_for_order(order)
-        find_setup_intent_for_order(order) || create_setup_intent_for_order(order)
-      end
-
-      def find_setup_intent_for_order(order)
-        SolidusStripe::SetupIntent
-          .find_by(order: order) # TODO: order.setup_intent (?)
-          &.stripe_setup_intent(self)
-      end
-
-      def create_setup_intent_for_order(order)
-        customer = customer_for(order.user) || customer_for_order(order)
-
-        intent = gateway.request do
-          Stripe::SetupIntent.create({
-            customer: customer,
-            usage: 'off_session', # TODO: use the payment method's preference
-            metadata: { solidus_order_number: order.number },
-          })
-        end
-
-        SolidusStripe::SetupIntent.create!(
-          order: order,
-          stripe_setup_intent_id: intent.id,
-        )
-
-        intent
-      end
-    end
-
-    concerning :PaymentIntents do
-      def find_or_create_payment_intent_for_order(order)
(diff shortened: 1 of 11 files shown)


lib/solidus_stripe/testing_support/factories.rb
```diff
@@ -17,4 +17,16 @@
     association :payment_method, factory: :stripe_payment_method
     stripe_payment_method_id { "pm_#{SecureRandom.uuid.delete('-')}" }
   end
+
+  factory :stripe_payment_intent, class: 'SolidusStripe::PaymentIntent' do
+    association :order
+    association :payment_method, factory: :stripe_payment_method
```
waiting-for-dev, line 23: I think at the moment of this commit, the payment_method association hasn't been added yet.

lib/solidus_stripe/testing_support/factories.rb
```diff
@@ -17,4 +17,16 @@
...
+    stripe_payment_intent_id { "pm_#{SecureRandom.uuid.delete('-')}" }
+  end
+
+  factory :stripe_setup_intent, class: 'SolidusStripe::SetupIntent' do
+    association :order
+    association :payment_method, factory: :stripe_payment_method
```
waiting-for-dev, line 29: Same here about the association not being present.

app/models/solidus_stripe/payment_method.rb
```diff
@@ -163,13 +163,18 @@ def payment_profiles_supported?
       false
     end
 
-    def stripe_dashboard_url(payment)
-      path_prefix = '/test' if preferred_test_mode
+    # Fetches the payment intent when available, falls back on the setup intent associated to the order.
+    def self.intent_id_for_payment(payment)
```
waiting-for-dev, line 96: I'm a bit confused. I don't see this method being called anywhere :thinking:

lib/solidus_stripe/testing_support/factories.rb
```diff
@@ -20,13 +20,11 @@
 
   factory :stripe_payment_intent, class: 'SolidusStripe::PaymentIntent' do
     association :order
-    association :payment_method, factory: :stripe_payment_method
```
waiting-for-dev: I think this is a consequence of the previous commit being in the wrong place.

app/models/solidus_stripe/payment_method.rb
```diff
@@ -163,13 +163,18 @@ def payment_profiles_supported?
...
+      return unless payment
 
-      intent_id =
-        payment.transaction_id ||
+      payment.transaction_id || (
```
waiting-for-dev: Checking it makes understanding the flow a bit confusing, as it looks like a third alternative. However, if I understand well, that should always be present in any case (although I think now it is only populated when using the payment intent flow). Thoughts on only using `payment.transaction_id` or only using the other two options?

app/models/solidus_stripe/setup_intent.rb
```diff
@@ -3,11 +3,26 @@
 module SolidusStripe
   class SetupIntent < Spree::Base
     belongs_to :order, class_name: 'Spree::Order'
+    belongs_to :payment_method, class_name: 'SolidusStripe::PaymentMethod'
 
-    def stripe_setup_intent(payment_method)
+    before_create :create_stripe_intent, unless: :stripe_setup_intent_id?
```
waiting-for-dev: Aren't we creating it explicitly?

waiting-for-dev (approved):
Thanks, @elia & @kennyadsl. Super happy seeing it taking shape. I'm approving to not block moving forward, but I left some comments I'd like feedback to better understanding :pray:

app/models/solidus_stripe/payment_method.rb
waiting-for-dev: However, the method will be very useful from the webhook handling part.

app/models/solidus_stripe/payment_method.rb
kennyadsl: The transaction_id will be populated in both cases but with SetupIntent, after the payment step and before confirming the order, the payment does not have a PaymentIntent nor a transaction_id yet. If we need to retrieve the intent at that point (eg. to show a link to stripe in the admin for that incomplete order), we have no way for doing that. 

That said, we know that `SolidusStripe::PaymentIntent.where(order: payment.order, payment_method: payment.payment_method)&.pick(:stripe_payment_intent_id)` is probably useless because it should always be the same as the `transaction_id`, but maybe there are some edge cases we are not considering and for now it might have sense to keep it. Open to alternatives!

app/models/solidus_stripe/setup_intent.rb
kennyadsl: Now that we removed the call to the custom `find_or_create_setup_intent_for_order` from the payment method, we are not creating this explicitly. Instead, we are relying on the AR interface in order to call `find_or_create_by`. They do the same thing, it's just more Rails way, and moved to the right place. 


Also, to be clear, we are creating the stripe intent in Stripe, and just saving its id on the internal intent. Do you think it's a matter of finding a better name?

app/models/solidus_stripe/payment_method.rb
kennyadsl: I think you are right. It's related to the dashboard_url. We probably need to update some partial to use this. Also, seems like that part is not tested yet.

lib/solidus_stripe/testing_support/factories.rb
```diff
@@ -15,6 +15,6 @@
 
   factory :stripe_payment_source, class: 'SolidusStripe::PaymentSource' do
     association :payment_method, factory: :stripe_payment_method
-    stripe_payment_method_id { 'pm_123' }
+    stripe_payment_method_id { "pm_#{SecureRandom.uuid.delete('-')}" }
```
kennyadsl, line 18: This was needed to avoid false positive specs. If all the generated ids have the same value, and we run checks against them, it's very likely that we could target the right object and struggle to realize that.

app/models/solidus_stripe/setup_intent.rb
waiting-for-dev: The issue is that, in my view, mixing persistence with network calls is not a good idea. I'd prefer a clear and explicit interface. Otherwise, we risk not handling network errors and leaving the db on an inconsistent state, besides probably slowing down some unit tests in the future.

kennyadsl: That's a very valid point. About slowing down testing, I think it won't happen if we have the factory setting the `intent_id` but it could happen anyway in some cases, where we are not explicitly using factories when creating intents.

In practice, do you suggest to explicitly doing both things from the caller method, defining a class method on the Setup/PaymentIntent that takes care of both things, or creating a service object? I'm fine with all of them, and also with a temporary solution, because maybe we will have the perfect place for this code after the work we are doing these days.

app/models/solidus_stripe/payment_method.rb
elia: We added a TODO to remind us that we should re-evaluate the implementation of the method before releasing

app/models/solidus_stripe/setup_intent.rb
elia: We (me and @waiting-for-dev) rearranged the code to avoid the AR callback and also avoid AR operations in the view 🙌

PR #216, opened 2023-03-14
Add SolidusStripe::Customer

### LINKED ISSUE #163: Track Stripe Customers with a `SolidusStripe::Customer` model
## What

Add a `SolidusStripe::Customer` model that can map `Spree::User` to `Stripe::Customer`, scoped to a `SolidusStripe::PaymentMethod`.

Whenever we create a PaymentIntent it should be associated to the correct customer id, which should be one for a single user. Payments done by guest users should not share the customer id (or not set it at all). 

Support for this with reusable payment sources will be tackled in a different issue.

### Open questions

Consider avoiding to expose the user id to stripe, and implications of reused user ids in development and testing.

```rb
           Stripe::Customer.create(
             email: user.email,
             metadata: { solidus_user_id: user.id },
           )
```

### Dependencies

- blocked by #154 

## Why

Payments should be tied to a single customer on stripe as they're coming from a single user on solidus, this will make handling issues and investigations much easier (e.g. from customer support team members).

Since we don't have an official way to associate a customer profile id (from the payment provider) to a solidus user, we need to map this somehow. The above implementation is suggested with a reference to a similar solution implemented in solidus_paypal_braintree.

elia:
## Summary

Fixes #163

Adds a `SolidusStripe::Customer` class that connects the user or the order (for guest users) to a stripe customer, ensuring it stays the same across order and intents generation.

- Use a polymorphic association to allow connecting it to either users or orders (for guest users)
- Stop using metadata since we now store the id and we can rely on the database for fetching it
- No need for fetching the customer form the API anywhere in the current implementation

## Checklist

### DIFF
--- a/db/migrate/20230313150008_create_solidus_stripe_customers.rb
+++ b/db/migrate/20230313150008_create_solidus_stripe_customers.rb
@@ -1,0 +1,16 @@
+class CreateSolidusStripeCustomers < ActiveRecord::Migration[7.0]
+  def change
+    create_table :solidus_stripe_customers do |t|
+      t.references :payment_method, null: false, foreign_key: { to_table: :spree_payment_methods }
+
+      t.string :source_type
+      t.integer :source_id
+
+      t.string :stripe_id, index: true
+
+      t.timestamps
+
+      t.index [:payment_method_id, :source_type, :source_id], unique: true, name: :payment_method_and_source
+    end
+  end
+end
--- a/app/models/solidus_stripe/customer.rb
+++ b/app/models/solidus_stripe/customer.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module SolidusStripe
+  class Customer < ApplicationRecord
+    belongs_to :payment_method
+
+    # Source is supposed to be a user or an order and needs to respond to #email
+    belongs_to :source, polymorphic: true
+
+    def self.retrieve_or_create_stripe_customer_id(payment_method:, order:)
+      instance = find_or_initialize_by(payment_method: payment_method, source: order.user || order)
+
+      instance.stripe_id ||
+        instance.create_stripe_customer.tap { instance.update!(stripe_id: _1.id) }.id
+    end
+
+    def create_stripe_customer
+      payment_method.gateway.request { Stripe::Customer.create(email: source.email) }
+    end
+  end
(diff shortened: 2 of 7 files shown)


kennyadsl (approved):

db/migrate/20230313150008_create_solidus_stripe_customers.rb
```diff
@@ -0,0 +1,14 @@
+class CreateSolidusStripeCustomers < ActiveRecord::Migration[7.0]
+  def change
+    create_table :solidus_stripe_customers do |t|
+      t.references :payment_method, null: false, foreign_key: { to_table: :spree_payment_methods }
+
+      t.string :source_type
+      t.integer :source_id
```
loicginoux, line 7: I would add a multi column index and for added safety make it unique.
```
add_index : solidus_stripe_customers, [: payment_method, : source_type, :source_id], unique: true
```

I can't remember if the `t.references :payment_method` creates automatically the index but it's not performant enough especially if the query you will do most of the time is the one in the customer model, generated by the line
```
 find_or_initialize_by(payment_method: payment_method, source: order.user || order)
```

It would be much more efficient to find existing ones if you add this index.

loicginoux: Also current code would not use it but still, for debugging or finding in console, I would also make a index on stripe_id column but that's more optionnal

waiting-for-dev (approved):

db/migrate/20230313150008_create_solidus_stripe_customers.rb
waiting-for-dev: Makes sense...

waiting-for-dev: It'll make writes more expensive, but I agree it's better to default to the safe side

kennyadsl: Good point; I'm fine addressing this here or creating a new issue for this optimization addressable later. Up to you @elia!

elia: Thanks for the feedback everyone, I'll add the index shortly, definitely an oversight not adding it 🙌

PR #217, opened 2023-03-14
Always create Solidus counter-part when creating Stripe payment intent

### LINKED ISSUE #218: Have all created payment intents associated to a Solidus payment intent record
## What

We're only creating a Solidus payment intent record for now when we go through the payment intent flow and render the payment element. We're not doing so when reusing a payment source or going through the setup intent flow. We want to create it in every situation.

## Why

When getting a webhook for a payment intent event, we need to have a reference to fetch the associated order and payment. It's also good for consistency.

waiting-for-dev:
## Summary

That'll allow us always to fetch the associated order when receiving an incoming payment intent webhook.

Until now, we were only creating the Solidus counterpart when going through the payment intent flow and not reusing a source. As the payment made that way is already authorized, the `purchase` method is not reached, so we don't end up with two payment intent resources.

The refactor makes our limitation more evident, where we can capture only the whole order amount.

We're also splitting the find & create paths for the intent models methods. The idea is to call `.create_stripe_intent` and not `retrieve_or_create_stripe_intent` when we already know that the intent is not there (i.e., from the gateway).

Closes #218 

## Checklist

### DIFF
--- a/app/models/solidus_stripe/gateway.rb
+++ b/app/models/solidus_stripe/gateway.rb
@@ -85,17 +85,15 @@ module SolidusStripe
     end
 
     # Authorizes and captures a certain amount on the provided payment source.
-    def purchase(amount_in_cents, _source, options = {})
-      currency = options.fetch(:currency)
-
+    # @todo add support for purchasing custom amounts
+    def purchase(_amount_in_cents, source, options = {})
       # Charge the Customer instead of the card:
-      payment_intent = request do
-        Stripe::PaymentIntent.create({
-          amount: to_stripe_amount(amount_in_cents, currency),
-          currency: currency,
-          **options[:payment_intent_options].to_h
-        })
-      end
+      payment_intent =
+        SolidusStripe::PaymentIntent.create_stripe_intent(
+          payment_method: source.payment_method,
+          order: options[:originator].order,
+          stripe_intent_options: options[:payment_intent_options] || {}
+        )
 
       build_payment_log(
         success: true,
--- a/app/models/solidus_stripe/payment_intent.rb
+++ b/app/models/solidus_stripe/payment_intent.rb
@@ -5,14 +5,13 @@ module SolidusStripe
     belongs_to :order, class_name: 'Spree::Order'
     belongs_to :payment_method, class_name: 'SolidusStripe::PaymentMethod'
 
-    def self.retrieve_or_create_stripe_intent(payment_method:, order:)
-      instance = find_or_initialize_by(payment_method: payment_method, order: order)
+    def self.retrieve_stripe_intent(payment_method:, order:)
+      find_by(payment_method: payment_method, order: order)&.stripe_intent
+    end
 
-      if instance.stripe_intent_id
-        instance.stripe_intent
-      else
-        instance.create_stripe_intent.tap { instance.update!(stripe_intent_id: _1.id) }
-      end
+    def self.create_stripe_intent(payment_method:, order:, stripe_intent_options: {})
+      instance = new(payment_method: payment_method, order: order)
+      instance.create_stripe_intent(stripe_intent_options).tap { instance.update!(stripe_intent_id: _1.id) }
     end
 
     def stripe_intent
@@ -21,7 +20,7 @@ module SolidusStripe
       end
     end
 
-    def create_stripe_intent
+    def create_stripe_intent(stripe_intent_options)
       customer = payment_method.customer_for(order)
 
       payment_method.gateway.request do
@@ -38,7 +37,7 @@ module SolidusStripe
           setup_future_usage: payment_method.preferred_setup_future_usage.presence,
           customer: customer,
           metadata: { solidus_order_number: order.number },
-        })
+        }.merge(stripe_intent_options))
       end
     end
   end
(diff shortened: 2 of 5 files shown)


kennyadsl:
I like this, we have a single point to create payment and setup intents. Still, we have other direct calls to the Stripe API in the Gateway class, which I'd love to avoid at this point. But it's one step further for sure so I'm ok addressing this later.

kennyadsl (approved):
👍 I'm fine with this, thanks @waiting-for-dev!

app/models/solidus_stripe/gateway.rb
```diff
@@ -85,17 +85,15 @@ def capture(amount_in_cents, payment_intent_id, options = {})
...
     # Authorizes and captures a certain amount on the provided payment source.
-    def purchase(amount_in_cents, _source, options = {})
-      currency = options.fetch(:currency)
-
+    # @todo add support for purchasing custom amounts
+    def purchase(_amount_in_cents, source, options = {})
       # Charge the Customer instead of the card:
-      payment_intent = request do
-        Stripe::PaymentIntent.create({
-          amount: to_stripe_amount(amount_in_cents, currency),
```
elia, line 94: The amount should stay there, this is the only use-case in which custom amounts are not an issue since we create the intent and charge the amount all in the same place.

app/models/solidus_stripe/gateway.rb
```diff
@@ -85,17 +85,15 @@ def capture(amount_in_cents, payment_intent_id, options = {})
...
-          currency: currency,
-          **options[:payment_intent_options].to_h
-        })
-      end
+      payment_intent =
+        SolidusStripe::PaymentIntent.create_stripe_intent(
+          payment_method: source.payment_method,
+          order: Spree::Order.find(options[:originator].order_id),
```
elia: Thoughts on doing `options[:originator].order` here?

elia: Needing to provide order and payment method to `create_stripe_intent` in the gateway feel a little bit like broken encapsulation, as it sits (at least in theory) at a lower level. Should this method live on the payment method itself and we merge the gateway to it instead? Thoughts?

elia (changes requested):
🙌 

I'm not sure about the gain of splitting the `find_or_create` into `find || create` but happy to merge anyway, especially if it's useful later on. 

I left a couple of comments/questions.

app/models/solidus_stripe/gateway.rb
```diff
@@ -85,17 +85,15 @@ def capture(amount_in_cents, payment_intent_id, options = {})
...
       # Charge the Customer instead of the card:
-      payment_intent = request do
-        Stripe::PaymentIntent.create({
-          amount: to_stripe_amount(amount_in_cents, currency),
-          currency: currency,
-          **options[:payment_intent_options].to_h
-        })
-      end
+      payment_intent =
+        SolidusStripe::PaymentIntent.create_stripe_intent(
```
kennyadsl, line 92: > I'm not sure about the gain of splitting the `find_or_create` into `find || create`

@elia how can we call `create_stripe_intent` here without splitting? I mean, we can but already knowing the find part won't retrieve anything and it will go in the create path of the method. Is that what you are proposing?

app/models/solidus_stripe/gateway.rb
waiting-for-dev: But how can we do that? I mean, now the amount is taken from the order.

app/models/solidus_stripe/gateway.rb
waiting-for-dev: Oh, yeah, makes sense :sweat_smile: Changed. Thanks for pointing that out.

waiting-for-dev: Yeah, I totally agree. Maybe that could be fixed when we move all API calls to the gateway or an adapter class?

app/models/solidus_stripe/gateway.rb
waiting-for-dev: We could still have the `retrieve_or_create...` method and use it only when rendering the Stripe element, while using the new `create...` from the gateway (otherwise the code reader might be confused wondering in which situation a payment intent could already be present). However, I'm not sure that we should make it for a single-use case.

waiting-for-dev: Happy to change it, though, if you think that's a good idea.

elia: Sorry, the comment was not clear 🫢 

I was referring mainly to the _stripe partial, where we don't know if there's one already or not, but it's fine, since we need the split for other things. All is good on that side 🙌

app/models/solidus_stripe/gateway.rb
elia: The gateway or the payment method are fine (solidus calls the methods on the payment method and are forwarded to the gateway via `delegate`). 

I don't think we need anything fancier than that as Stripe is pretty easy to intercept for testing and probably that's the only place in which you would want to switch implementation, I guess 😊.

app/models/solidus_stripe/gateway.rb
elia: Oh, right, I was confused by the namespace, I get it now. This goes back to the encapsulation being a bit off but we can fix it in another iteration 👍

elia (approved):
Thanks for the quick turnaround, we definitely need to rethink the encapsulation, but for now we can merge and unlock the remaining work on webhooks ✅

PR #222, opened 2023-03-16
Handle payment_intent.succeeded event

### LINKED ISSUE #179: Handle `payment_intent.succeeded` webhook event

waiting-for-dev:
## Summary

A `payment_intent.successful` event [is published](https://stripe.com/docs/payments/payment-intents/verifying-status#webhooks) when capturing a payment intent from the Stripe dashboard. At that point, we need to update the associated Solidus payment to be marked as completed.

However, the webhook could also be received after capturing the payment from the Solidus admin interface, as it updates through Stripe's API. Therefore, we need to do nothing if the payment is already completed.

The unit tests on the subscriber cover the handler's behavior, while at the integration layer, we're making sure that it's actually reached when the "payment_intent.succeeded" event is published. Besides, this PR has been manually tested for different checkout flows.

Closes #179

## Checklist

### DIFF
--- a/app/subscribers/solidus_stripe/webhook/payment_intent_subscriber.rb
+++ b/app/subscribers/solidus_stripe/webhook/payment_intent_subscriber.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+module SolidusStripe
+  module Webhook
+    # Handlers for Stripe payment_intent events.
+    class PaymentIntentSubscriber
+      include Omnes::Subscriber
+
+      handle :"stripe.payment_intent.succeeded", with: :complete_payment
+
+      # Captures a payment.
+      #
+      # Marks a Solidus payment associated to a Stripe payment intent as
+      # completed, adding a log entry about the event.
+      #
+      # @param event [SolidusStripe::Webhook::Event]
+      def complete_payment(event)
+        payment_intent_id = event.data.object.id
+        payment = Spree::Payment.find_by!(response_code: payment_intent_id)
+        return if payment.completed?
+
+        payment.complete!.tap do
+          SolidusStripe::LogEntries.payment_log(
+            payment,
+            success: true,
+            message: "Capture was successful after payment_intent.succeeded webhook"
+          )
+        end
+      end
+    end
+  end
+end
--- a/spec/support/solidus_stripe/webhook/event_with_context_factory.rb
+++ b/spec/support/solidus_stripe/webhook/event_with_context_factory.rb
@@ -8,7 +8,8 @@ module SolidusStripe
     # Factory to create a webhook event along with its context.
     #
     # It allows to create Stripe webhook from different sources (hash, stripe
-    # object) in different representations (json, stripe event object, header).
+    # object) in different representations (json, stripe event object, solidus
+    # stripe object, header).
     #
     # The context for a event is composed by the timestamp and its secret, which
     # in turn affect the header representation.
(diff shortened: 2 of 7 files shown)


app/subscribers/solidus_stripe/webhook/payment_intent_subscriber.rb
```diff
@@ -0,0 +1,24 @@
...
+      # Completes a payment.
+      #
+      # Marks a Solidus payment associated to a Stripe payment intent as
+      # completed.
+      #
+      # @param event [SolidusStripe::Webhook::Event]
+      def complete_payment(event)
+        payment_intent_id = event.data.object.id
+        payment = Spree::Payment.find_by!(response_code: payment_intent_id)
+        payment.complete! unless payment.completed?
```
rainerdema: After the event has been processed, how about saving a `SolidusStripe::LogEntries.payment_log` related to the event on the payment?
By doing so, it would enable the admin side to display a timeline of events related to the payment, allowing for a clearer understanding of why, in this case, the payment is marked as completed 🤔 WDYT?

spec/support/solidus_stripe/webhook/event_with_context_factory.rb
```diff
@@ -56,6 +57,11 @@ def stripe_object
         @stripe_object ||= Stripe::Event.construct_from(data)
       end
 
+      def solidus_stripe_object
```
kennyadsl, line 60: Can you please help me understand why this method is called `solidus_stripe_object` and returns a Webhook Event?

rainerdema (commented):
Looks great! Just a small consideration 👀

kennyadsl (approved):
Nice and simple. I left a question about a method name that I don't fully understand, but it's most likely me.

app/subscribers/solidus_stripe/webhook/payment_intent_subscriber.rb
waiting-for-dev: Makes total sense. I just added that. Thanks for your suggestion :+1:

spec/support/solidus_stripe/webhook/event_with_context_factory.rb
waiting-for-dev: The `EventWithContext` class can return a Stripe event in different representations:

- `#json`: A JSON string representation
- `#signature_header`: How it looks signed in the request header
- `#stripe_object`: As a `Stripe::Event` object.
- `#solidus_stripe_object`: As a `SolidusStripe::Webhook::Event` object.

I know it's a bit cryptic, but I can't find better candidates. We can't call the `#stripe_object` method `#stripe_event`, because all those methods return representations of a Stripe event. Once we have that method, the most logical candidate for the new addition is `#solidus_stripe_object`.

Happy to change it if we find a better solution!!

rainerdema (approved):
🚢

elia (approved):

Issue #224, opened 2023-03-17
Error when completing an order with a potential fraudulent payment


waiting-for-dev:
**Steps to reproduce**

- Perform a checkout and use [card number 4100000000000019](https://stripe.com/docs/testing#fraud-prevention) to pay with Stripe.
- The order advances successfully from the payment to the confirm step. However, when clicking on the "Complete" button (i.e., when Stripe tries to authorize the payment intent), a `Stripe::CardError` is raised.

[screencast-localhost_3000-2023.03.17-09_27_55.webm](https://user-images.githubusercontent.com/52650/225869555-47d6b55a-cc4f-454d-befa-04c68f7ab094.mp4)

rainerdema:
After more investigation, I've found that the following cards fail with the same exception:
| Number| Description | Source |
|--------|--------|------------|
| 4100000000000019 | The charge has a [risk level of “highest”](https://stripe.com/docs/radar/risk-evaluation#high-risk). Radar always blocks it.| [Stripe Fraud Prevention Testing](https://stripe.com/docs/testing?testing-method=card-numbers#fraud-prevention) |
| 4000000000004954 | The charge has a [risk level of “highest”](https://stripe.com/docs/radar/risk-evaluation#high-risk). Radar might block it [depending on your settings](https://stripe.com/docs/radar/risk-settings). | [Stripe Fraud Prevention Testing](https://stripe.com/docs/testing?testing-method=card-numbers#fraud-prevention) |
| 4000000000000341 | Attaching this card to a [Customer](https://stripe.com/docs/api/customers) object succeeds, but attempts to charge the customer fail. | [Stripe Declined Payments Testing](https://stripe.com/docs/testing?testing-method=card-numbers#declined-payments) |

Note: In my case, these only fail with setup intent configured during the payment intent creation.

PR #233, opened 2023-03-20
Fix `Gateway#purchase` not capturing the payment

### LINKED ISSUE #232: Change `Gateway#purchase` to both authorize and capture a payment
On Solidus, `Gateway#purchase` is meant to both authorize and capture a payment. However, we're only doing the first part, and the method actually contains what should be the `#authorize` method.

waiting-for-dev:
## Summary

On Solidus, `Gateway#purchase` is meant to authorize and capture a payment in a single step. However, we were using that method only to authorize (confirm, on Stripe) the payment.

We're now moving the authorize logic to a helper method shared by the `#authorize` and `#purchase` methods, while we use the `automatic` capture method on `#purchase`.

We're also unifying the format for the success messages recorded in the response.

We're getting rid of the option to pass custom payment intent options on `#authorize`. That was the only method in the gateway supporting them, and we can add it later to all the methods if we feel that's needed.

Closes #232

We take the occasion to perform some house-keeping (two last commits):

- Extract and reuse checks on the Gateway method arguments. We were only performing safety checks against `Gateway#capture` arguments. However, they also make sense for the other methods in the class. We're also adding unit tests for those checks.
- Unify naming convention for variables in the gateway tests

## Checklist

### DIFF
--- a/app/models/solidus_stripe/gateway.rb
+++ b/app/models/solidus_stripe/gateway.rb
@@ -32,16 +32,21 @@ module SolidusStripe
     # Authorizes a certain amount on the provided payment source.
     #
     # @see #purchase
-    def authorize(amount_in_cents, source, options = {})
-      stripe_payment_method = source.stripe_payment_method
-
-      payment_intent_options = options[:payment_intent_options].dup.to_h
-      payment_intent_options[:capture_method] = "manual"
-      payment_intent_options[:confirm] = true
-      payment_intent_options[:payment_method] = stripe_payment_method.id
-      payment_intent_options[:customer] = stripe_payment_method.customer
-
-      purchase(amount_in_cents, source, options.merge(payment_intent_options: payment_intent_options))
+    def authorize(_amount_in_cents, source, options = {})
+      stripe_payment_intent = create_stripe_payment_intent(
+        source,
+        options[:originator].order,
+        stripe_payment_intent_options: {
+          capture_method: "manual",
+          confirm: true
+        }
+      )
+      build_payment_log(
+        success: true,
+        message: "PaymentIntent was confirmed successfully",
+        response_code: stripe_payment_intent.id,
+        data: stripe_payment_intent,
+      )
     end
 
     # Captures a certain amount from a previously authorized transaction.
@@ -68,11 +73,10 @@ module SolidusStripe
         raise ArgumentError, "the payment-intent id has the wrong format"
       end
 
-      payment_intent = request { Stripe::PaymentIntent.capture(payment_intent_id) }
-
+      payment_intent = capture_stripe_payment_intent(payment_intent_id)
       build_payment_log(
         success: true,
-        message: "Capture was successful",
+        message: "PaymentIntent was confirmed successfully",
         response_code: payment_intent.id,
         data: payment_intent,
       )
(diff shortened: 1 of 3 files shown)


app/models/solidus_stripe/gateway.rb
```diff
@@ -87,19 +91,20 @@ def capture(amount_in_cents, payment_intent_id, options = {})
...
-
+      stripe_payment_intent = create_stripe_payment_intent(
+        source,
+        options[:originator].order,
+        stripe_payment_intent_options: {
+          capture_method: "manual",
+          confirm: true
+        }
+      )
+      captured_stripe_payment_intent = capture_stripe_payment_intent(stripe_payment_intent.id)
```
elia: It seems to me from the wording of the commit description that we're assuming that this method was not doing both authorization and capture before this change. 

Otherwise I'm not sure I understand why we're now doing the capture separately instead of at once with the default `capture_method`, which is _automatic_.

https://stripe.com/docs/api/payment_intents/create#create_payment_intent-capture_method

app/models/solidus_stripe/gateway.rb
```diff
@@ -182,5 +176,26 @@ def create_stripe_payment_intent(source, order, stripe_payment_intent_options:)
     def capture_stripe_payment_intent(stripe_payment_intent_id)
       request { Stripe::PaymentIntent.capture(stripe_payment_intent_id) }
     end
+
+    def check_given_amount_matches_order_total(amount_in_cents, options)
```
elia: There's no mention of the order here, thoughts on just stating that we _check given amount against the payment intent amount_?

elia (commented):
Happy to see the gateway cleaned up 🙏 

Left a question plus a nit

app/models/solidus_stripe/gateway.rb
```diff
@@ -51,28 +58,13 @@ def authorize(amount_in_cents, source, options = {})
...
+      check_payment_intent_id(payment_intent_id)
 
+      payment_intent = capture_stripe_payment_intent(payment_intent_id)
       build_payment_log(
         success: true,
-        message: "Capture was successful",
+        message: "PaymentIntent was confirmed successfully",
         response_code: payment_intent.id,
         data: payment_intent,
       )
```
rainerdema, line 70: Not strictly related, but I was thinking about what happens in case of exceptions, and I noticed that we're logging a successful log with `build_payment_log(success: true, ...)`, but there might be cases where `capture_stripe_payment_intent` raises an error, and we handle only `Stripe::InvalidRequestError`.

Are we sure that this is the only exception to handle? I was looking for other possible exceptions, and maybe we can handle a more generic [`Stripe::StripeError`](https://github.com/stripe/stripe-ruby/blob/980b7b011f6fbff276ecadce5dd884e00669724f/lib/stripe/errors.rb#L6) at all the points when we're doing these external calls 🤔 WDYT?

If you believe this concern is not directly related to the changes proposed in this PR, we can discuss it at a later time 👍

app/models/solidus_stripe/gateway.rb
waiting-for-dev: I overlooked that! Thanks, I updated the code.

app/models/solidus_stripe/gateway.rb
waiting-for-dev: Makes sense. Changed!

rainerdema (approved):
Changes looks good, and I don't have much more to add. Great job! 👏

elia (approved):
🙌

waiting-for-dev:
Rebasing on top of master after merging https://github.com/solidusio/solidus_stripe/pull/235 to fix CI

app/models/solidus_stripe/gateway.rb
waiting-for-dev: @rainerdema, yeah, it looks like, for now, only the happy path is covered in some situations, and I'm also unsure whether `InvalidRequestError` handles all possible failures. That's definitely food for another PR (probably related to #224)

Issue #185, opened 2023-02-23
Update Stripe when a payment is failed on Solidus part


waiting-for-dev:
**Steps to reproduce**

- Proceed with a checkout until Stripe's payment form is rendered. At that point, Solidus has already created a `:processing` payment associated with the order (as we need the payment, which is bound to the Payment Intent, to render Stripe's Payment Element).
- Mark the payment as failed in the backend.
- Submit the form from the checkout with a valid card.

**Result**
Stripe authorizes the payment, while Solidus shows an error message because of trying to move a failed payment forward.

**Expected result**
Stripe has already marked that payment as invalid and we get the appropriate response.

**Considerations**
 It's not a very likely scenario, as just after the form is submitted in a normal flow, the payment will be marked as `:pending`, and the transition to `:failure` won't be possible anymore.

rainerdema:
With the introduction of #249, we will no longer have a payment in the `payment` step but in the `confirm` one.
This payment will be in `checkout` state with no available actions for the admin.

https://user-images.githubusercontent.com/19948291/228614814-39bdf836-d42f-4f65-99b1-378740e8630d.mov

waiting-for-dev:
Cool, @rainerdema! Does it mean we can close this one?

rainerdema:
> Cool, @rainerdema! Does it mean we can close this one?

Yeah I think so 👍 I'm waiting to merge #249 before closing this one

PR #246, opened 2023-03-22
Handle payment_intent.failed event

### LINKED ISSUE #179: Handle `payment_intent.succeeded` webhook event

waiting-for-dev:
## Summary

A `payment_intent.failed` event is [published](https://stripe.com/docs/payments/payment-intents/verifying-status#webhooks) when a payment intent can't be captured from Stripe.

After we check #185, the webhook could also be triggered by an action from the backend, so we need to do nothing if the payment is already failed.

Closes #179

## Checklist

### DIFF
--- a/app/subscribers/solidus_stripe/webhook/payment_intent_subscriber.rb
+++ b/app/subscribers/solidus_stripe/webhook/payment_intent_subscriber.rb
@@ -7,6 +7,7 @@ module SolidusStripe
       include Omnes::Subscriber
 
       handle :"stripe.payment_intent.succeeded", with: :complete_payment
+      handle :"stripe.payment_intent.payment_failed", with: :fail_payment
 
       # Captures a payment.
       #
@@ -15,8 +16,7 @@ module SolidusStripe
       #
       # @param event [SolidusStripe::Webhook::Event]
       def complete_payment(event)
-        payment_intent_id = event.data.object.id
-        payment = Spree::Payment.find_by!(response_code: payment_intent_id)
+        payment = extract_payment_from_event(event)
         return if payment.completed?
 
         payment.complete!.tap do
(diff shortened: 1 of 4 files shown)


kennyadsl (approved):
👍

app/subscribers/solidus_stripe/webhook/payment_intent_subscriber.rb
```diff
@@ -27,6 +27,32 @@ def complete_payment(event)
...
+
+      # Voids a payment.
+      #
+      # Marks a Solidus payment associated to a Stripe payment intent as
+      # failed, adding a log entry about the event.
+      #
+      # @param event [SolidusStripe::Webhook::Event]
+      def fail_payment(event)
+        payment = extract_payment_from_event(event)
+        return if payment.failed?
```
loicginoux, line 39:
```suggestion
        return if !payment.can_failure?
```

I'd say that checking more generally if we can transition to failed is more generic and safe. You could be on a state (like checkout) where payment is not failed but doing the transition will raise an error. Same goes for the complete! event of the same file.

app/subscribers/solidus_stripe/webhook/payment_intent_subscriber.rb
```diff
@@ -27,6 +27,32 @@ def complete_payment(event)
...
+
+        payment.failure!.tap do
+          SolidusStripe::LogEntries.payment_log(
+            payment,
+            success: true,
```
elia: Shouldn't this be `false`?

waiting-for-dev: Oh, you're right. Changed! Thanks for pointing out.

I also added another commit to test the success value for both handlers in the subscriber class.

app/subscribers/solidus_stripe/webhook/payment_intent_subscriber.rb
waiting-for-dev: Thanks for your suggestion, @loicginoux. If Solidus payment is in the `:checkout` state, but we're already getting a `payment_intent.payment_failed` webhook, that's probably an abnormal situation. I think it's better to be stricter so we can't detect any failure in our integration.

PR #249, opened 2023-03-23
Implement support for deferred PaymenIntent confirmation

### LINKED ISSUE #185: Update Stripe when a payment is failed on Solidus part

### LINKED ISSUE #190: Remove payment handling code from the callback controller
Stripe recommends using webhooks as the source of truth, so let's just show a notice about the order being completed.

### LINKED ISSUE #209: Handle extra SCA request at the confirm step
## Why

Some cards/payment methods, especially in Europe, will require additional confirmation from the buyer when the card is charged or the amount authorized. We do that just before order completion for a number of cases (setup intents, saved payments).

## What

Check if there's some JS code we can embed in the confirmation or completion page or what is needed to ensure we allow buyers to take the necessary actions to confirm payment intents that end up in the `requires_action` status.

elia:
## Summary

Fixes #209, Fixes #185, Fixes #190

This PR improves the payment process by replacing the double flow of `Payment Intent`/`Setup Intent` with a `Payment Intent` flow in a specific scenario described in the Stripe documentation:
[Build a two-step confirmation experience
](https://stripe.com/docs/payments/build-a-two-step-confirmation)

With the new approach, the two-step subdivision achieves the following issues:
- The creation of the Stripe payment method remains in the payment step (as before)
- The Payment intent creation will be invoked during the confirm step rendering.
- The payment authorization occurs when the order is submitted on the confirm page.

The benefits of this update include:
- Removing compromises in the payment intent flow that previously required "skipping" the confirm step due to the premature authorization executed in the payment step.
- Aligning the authorization process with the Solidus checkout flow.
- Removing the `authorize` and `purchase` methods from the gateway, as they are no longer needed.

## Checklist


db/migrate/20230323154931_drop_solidus_stripe_setup_intent.rb
```diff
@@ -1,5 +1,13 @@
...
-    drop_table :solidus_stripe_setup_intents
+    drop_table "solidus_stripe_setup_intents", force: :cascade do |t|
+      t.string "stripe_intent_id"
+      t.integer "order_id", null: false
+      t.datetime "created_at", null: false
+      t.datetime "updated_at", null: false
+      t.integer "payment_method_id", null: false
+      t.index ["order_id"], name: "index_solidus_stripe_setup_intents_on_order_id"
+      t.index ["payment_method_id"], name: "index_solidus_stripe_setup_intents_on_payment_method_id"
+    end
```
waiting-for-dev, line 11: Nit: that belongs to the previous commit

waiting-for-dev (dismissed):
So much simpler!

app/models/solidus_stripe/payment_intent.rb
```diff
@@ -5,41 +5,69 @@ class PaymentIntent < ApplicationRecord
...
+      successful
+    end
+
+    def stripe_intent
+      @stripe_intent ||= payment_method.gateway.request do
+        Stripe::PaymentIntent.retrieve(stripe_intent_id)
+      end
+    end
+
+    def reload
```
kennyadsl: No one is using this method in this commit

kennyadsl (commented):
Maybe spotted a method that we are not using, other than that, 👍

app/models/solidus_stripe/payment_intent.rb
rainerdema: This was added because we thought it could be useful in the future in case we want to refresh a Payment Intent from Stripe.

Maybe we can move it to a separate commit?

kennyadsl: I'd say that we can add it when we need it, even if it's tested, it's still another public method that we need to maintain.

app/controllers/solidus_stripe/intents_controller.rb
```diff
@@ -8,59 +8,34 @@ class SolidusStripe::IntentsController < Spree::BaseController
...
-    else
+    unless params[:payment_intent]
       return head :unprocessable_entity
     end
 
-    payment = intent.create_payment!(
-      amount: current_order.total, # TODO: double check, remove store credit?
-      add_to_wallet: true
-    )
+    unless %w[confirm].include?(current_order.state.to_s)
```
waiting-for-dev, line 15:
```suggestion
    unless current_order.state.to_s == "confirm"
```

waiting-for-dev:
```suggestion
    unless current_order.confirm?
```

app/models/solidus_stripe/gateway.rb
```diff
@@ -29,38 +29,6 @@ def initialize(options)
...
 
-    # Authorizes a certain amount on the provided payment source.
-    #
-    # We create and confirm the Stripe payment intent in two steps. That's to
-    # guarantee that we associate a Solidus payment on the creation, and we can
-    # fetch it after the webhook event is published on confirmation.
-    #
-    # The Stripe payment intent id is copied over the Solidus payment
-    # `response_code` field.
-    def authorize(amount_in_cents, source, options = {})
```
waiting-for-dev, line 40: I am concerned about removing the gateway methods. Are we sure there's no situation where someone could call `#authorize` from the backend? Also, they might be helpful as library code. Otherwise, we're coupling the implementation to SSF to a great extent.

waiting-for-dev (approved):
Thanks, @elia & @rainerdema! :heart: Left one concern about the removal of the gateway methods.

I'm also wondering if we need the `SolidusStripe::PaymentIntent` model at all. It's not needed from the webhook handlers, as the `response_code` is filled on payment intent creation. Can you help me understand? Is it only needed in the case a user refreshes the confirm page?

app/models/solidus_stripe/payment_intent.rb
elia: I see this case as similar to adding `method_missing` and not implementing `respond_to?`, I don't know if there's a better way or rails support to reset instance variables on reload that's definitely what we need if we use ivars inside an AR model.

Not much unlike what is done by AR itself (e.g. [here](https://github.com/rails/rails/blob/7c70791470fc517deb7c640bead9f1b47efb5539/activerecord/lib/active_record/autosave_association.rb#L233-L238)).

@kennyadsl I'm sort of ok with removing it but I'd like to hear your thoughts with the context provided above first 🙌

kennyadsl: Ah, I didn't think it's needed when we call the already existing `reload` method on the instance. If we need it to make the `reload` method work correctly for `PaymentIntent`, I definitely favor keeping it. Maybe it just belongs to a separate PR?

app/models/solidus_stripe/gateway.rb
elia: > Otherwise, we're coupling the implementation to SSF to a great extent.

I think this has more to do with how stripe works, which has the authorization done as part of the in-browser confirmation.

Any other use of the Stripe API should be accessed through the `stripe` gem, possibly using the credentials provided by the payment method with `payment_method.gateway.request {…}`.

@waiting-for-dev you see any other case in which having `#authorize` available in the gateway would be useful?

app/models/solidus_stripe/payment_intent.rb
kennyadsl: But if it's needed for something we added here, it's totally ok to have it. If that's the case, I think it's just not super clear, so maybe a couple of lines in the commit description would be good, or a separate commit.

elia: I'll ensure it's part of the commit introducing the memoization and properly explain in the commit description 👍

app/models/solidus_stripe/gateway.rb
kennyadsl: Yes, there's not a lot we can do here. We could keep these methods for those who want to use other flows (maybe some more elaborate wallet implementation with subscriptions?) but it's a bit speculating at this point.

waiting-for-dev: Ok, let's remove them, and we can always add them later if needed.

kennyadsl: That said, a noop with a comment (and maybe a simple implementation example) would not be bad!

elia: @kennyadsl even for a README entry I'd like to have a concrete use-case that we can use to exemplify, otherwise we should seriously consider that it's very unlikely that anyone will ever need it.

kennyadsl: Yes, the simple implementation example I meant was as a comment or documentation, being clear it's some form of pseudo-code. But maybe it's just a matter of linking the stripe documentation?

elia: @kennyadsl it's possible that authorize/purchase methods will be required by https://github.com/solidusio/solidus_stripe/issues/266.

Can we defer this discussion to that ticket? Is that fine?

kennyadsl: Sure, but can we at least open an issue so that if that's not covered by #266, we don't forget?

kennyadsl (approved):

app/models/solidus_stripe/gateway.rb
elia: Absolutely, here it is: https://github.com/solidusio/solidus_stripe/issues/267

elia:
> I'm also wondering if we need the SolidusStripe::PaymentIntent model at all. It's not needed from the webhook handlers, as the response_code is filled on payment intent creation. Can you help me understand? Is it only needed in the case a user refreshes the confirm page?

@waiting-for-dev sorry, I merged before replying to this question. And yes, it's there to ensure we have a single payment intent for each order. As you pointed out there might be indirect ways to do the same through `Payment#response_code` but as we discussed when we introduced the class this way it should be more reliable.

Issue #178, opened 2023-02-21
Support `auto_capture: true` for the payment method


elia:
## Why

Being able to auto_capture payments during the order completion is a stock solidus capability.

## What

Hook into the finalization process to capture the authorized amount for stripe payments that were previously authorized if `auto_capture` is enabled.

A possible implementation involves using a subscriber

```rb
# app/subscribers/update_account_balance_subscriber.rb
class SolidusStripe::AutoCaptureSubscriber
  include Omnes::Subscriber

  handle :order_finalized, with: :call

  def call(event)
    stripe_payments.each(&:capture!) # pseudo-code!
  end
end

# config/initializers/solidus_stripe.rb
Rails.application.config.to_prepare do
  SolidusStripe::AutoCaptureSubscriber.new.subscribe_to(Spree::Bus)
end
```

waiting-for-dev:
We'll need to implement something like https://github.com/solidusio/solidus_stripe/pull/245 for the unhappy path

rainerdema:
This seems to work fine with the current implementation.
Thanks to the implementation of the [`purchase` action with the "automatic" capture_method](https://github.com/solidusio/solidus_stripe/blob/6ecda67a456dd5d7ed08816b81470f98210c3ea8/app/models/solidus_stripe/gateway.rb#L89-L110), the authorization+capture happens correctly and is in line with the current Solidus logic.

As specified in the [`Spree::Payment::Processing` module](https://github.com/solidusio/solidus/blob/6d5580068d00c8fb711710a71e350e4c1358946c/core/app/models/spree/payment/processing.rb#L19-L21), if `auto_capture` is set to `true`, the purchase action will be invoked on the gateway of the referenced payment method.

I will do other checks, [follow Marc's suggestion about the unhappy path](https://github.com/solidusio/solidus_stripe/issues/178#issuecomment-1479410507) and implement the necessary specs to cover this use case.

waiting-for-dev:
Does it also work for the payment intent flow? As I understand, `manual` [is fixed](https://github.com/solidusio/solidus_stripe/blob/c4d127e1a52454700ec9fc180716ed9dec3eabe6/app/models/solidus_stripe/payment_intent.rb#L58) now.

PR #252, opened 2023-03-28
Prepare for deferred confirm

elia:
## Summary

Preparation for #249, these are all commits that are preparatory but not strictly related to the deferred confirm.

## Checklist

### DIFF
--- a/app/models/solidus_stripe/payment_intent.rb
+++ b/app/models/solidus_stripe/payment_intent.rb
@@ -52,10 +52,7 @@ module SolidusStripe
             order.currency,
           ),
           currency: order.currency,
-
-          # The capture method should stay manual in order to
-          # avoid capturing the money before the order is completed.
-          capture_method: 'manual',
+          capture_method: payment_method.auto_capture? ? 'automatic' : 'manual',
           setup_future_usage: payment_method.preferred_setup_future_usage.presence,
           customer: stripe_customer_id,
           metadata: { solidus_order_number: order.number },
--- a/lib/generators/solidus_stripe/install/install_generator.rb
+++ b/lib/generators/solidus_stripe/install/install_generator.rb
@@ -78,6 +78,18 @@ module SolidusStripe
               template engine.root.join(path), path
             end
           end
+
+          storefront_layout_path = Rails.root.join("app/views/layouts/storefront.html.erb")
+
+          if storefront_layout_path.exist?
+            insert_into_file(
+              storefront_layout_path.to_s,
+              "\n    <script async src='https://js.stripe.com/v3/'></script>",
+              before: %r{\s*</head>},
+            )
+          end
+
+          run 'bin/importmap pin @stripe/stripe-js'
         end
       end
 
--- a/lib/generators/solidus_stripe/install/templates/app/javascript/controllers/solidus_stripe_intent_controller.js
+++ b/lib/generators/solidus_stripe/install/templates/app/javascript/controllers/solidus_stripe_intent_controller.js
@@ -1,5 +1,5 @@
-import { Controller } from '@hotwired/stimulus'
-import { loadStripe } from '@stripe/stripe-js'
+import { Controller } from "@hotwired/stimulus"
+import { loadStripe } from "@stripe/stripe-js"
 
 export default class extends Controller {
   static values = {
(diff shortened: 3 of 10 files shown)


app/models/solidus_stripe/payment_intent.rb
```diff
@@ -52,10 +52,7 @@ def create_stripe_intent(stripe_intent_options)
             order.currency,
           ),
           currency: order.currency,
-
-          # The capture method should stay manual in order to
-          # avoid capturing the money before the order is completed.
-          capture_method: 'manual',
+          capture_method: payment_method.auto_capture? ? 'automatic' : 'manual',
```
waiting-for-dev, line 55: Does it fix #178? If so, do we need to implement the unhappy path (see https://github.com/solidusio/solidus_stripe/issues/178#issuecomment-1479410507). IMHO, if so, at least this one should deserve a separate PR for better reference in the future.

lib/generators/solidus_stripe/install/templates/app/javascript/controllers/solidus_stripe_intent_controller.js
```diff
@@ -32,6 +32,9 @@ export default class extends Controller {
 
   // @action
   async handleSubmit(e) {
+    // Bail out if not on the payment method form.
+    if (e.target !== this.radioOutletElement.form) return
```
waiting-for-dev, line 36: How could that happen??

waiting-for-dev (approved):
Thanks! Left two questions and a non-blocking suggestion :pray:

app/models/solidus_stripe/payment_intent.rb
kennyadsl: Agree with @waiting-for-dev here. I'm fine doing this before, and even partially (updating the issue with the remaining things to do), but I'd like this not to be part of the preparatory PR.

lib/generators/solidus_stripe/install/install_generator.rb
```diff
@@ -78,6 +78,18 @@ def install_solidus_starter_frontend_support
               template engine.root.join(path), path
             end
           end
+
+          storefront_layout_path = Rails.root.join("app/views/layouts/storefront.html.erb")
+
+          if storefront_layout_path.exist?
+            insert_into_file(
+              storefront_layout_path.to_s,
+              "\n    <script async src='https://js.stripe.com/v3/'></script>",
```
kennyadsl, line 87: Why this approach?

kennyadsl (commented):
Left a couple of comments/questions before approval. Overall I'm ok with this but the majority of the changes are not related to the preparation. 

As usual, I'd like to have multiple PRs (eg. for the sandbox stuff), but this is not blocking.

lib/generators/solidus_stripe/install/templates/app/javascript/controllers/solidus_stripe_intent_controller.js
elia: If you have any other form in the page that triggers a submit event (e.g. a subscribe to newsletter in the footer) this action will be triggered, since we listen for a generic `submit@window`. When we'll update the starter frontend with stimulus controllers and can be used as [outlets](https://stimulus.hotwired.dev/reference/outlets) we might be able to clean up this considerably.

app/models/solidus_stripe/payment_intent.rb
elia: This was added because the comment didn't seem relevant anymore with the deferred confirm and so it made sense to connect the value to the auto_capture setting instead of just deleting the comment. 
But open to postponing the change, it's not strictly required for the deferred capture PR.

lib/generators/solidus_stripe/install/install_generator.rb
elia: This comes straight outta `importmaps-rails` https://github.com/rails/importmap-rails/blob/main/lib/install/install.rb#L1-L9 and is in line with Stripe advising to load the JS on every page to better support fraud prevention. The stripe-js imported in .js files on the other hand is just a thin layer that will load this same file if it's not already available, so we can have the global script tag marked as `async` and at the same time use a comfy `import` in our JS files.

elia:
This was split in #253 #254 #255 #256

PR #255, opened 2023-03-29
Improve payment form handling and start using `@stripe/stripe-js` module

elia:
## Summary

Extracted from #249 

Use the [@stripe/stripe-js](https://www.npmjs.com/package/@stripe/stripe-js) module with `importmap`, as partially described in the [Stripe doc](https://stripe.com/docs/libraries/stripejs-esmodule#web-stripejs-esmodule).

The `<script>` tag is added to every page to support fraud detection as suggested in the stripe-js repo.
We opted for using the script tag instead of relying on the implicit side-effect because the browser will be able to start loading it earlier if it's directly in the HTML.

https://github.com/stripe/stripe-js#ensuring-stripejs-is-available-everywhere

## Checklist

### DIFF
--- a/lib/generators/solidus_stripe/install/install_generator.rb
+++ b/lib/generators/solidus_stripe/install/install_generator.rb
@@ -78,6 +78,8 @@ module SolidusStripe
               template engine.root.join(path), path
             end
           end
+
+          run 'bin/importmap pin @stripe/stripe-js'
         end
       end
 
--- a/lib/generators/solidus_stripe/install/templates/app/javascript/controllers/solidus_stripe_intent_controller.js
+++ b/lib/generators/solidus_stripe/install/templates/app/javascript/controllers/solidus_stripe_intent_controller.js
@@ -1,5 +1,5 @@
-import { Controller } from '@hotwired/stimulus'
-import { loadStripe } from '@stripe/stripe-js'
+import { Controller } from "@hotwired/stimulus"
+import { loadStripe } from "@stripe/stripe-js"
 
 export default class extends Controller {
   static values = {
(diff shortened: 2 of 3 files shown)


lib/generators/solidus_stripe/install/install_generator.rb
```diff
@@ -78,6 +78,18 @@ def install_solidus_starter_frontend_support
...
+
+          if storefront_layout_path.exist?
+            insert_into_file(
+              storefront_layout_path.to_s,
+              "\n    <script async src='https://js.stripe.com/v3/'></script>",
+              before: %r{\s*</head>},
+            )
+          end
+
+          run 'bin/importmap pin @stripe/stripe-js'
```
kennyadsl, line 82: In the documentation you linked, there are two alternative methods to load Stripe: manually and via a package manager. It seems to me that we are using both in this way. I think, but not 100% sure yet, that there's way to just use importmaps to load stripe code (eg, via https://unpkg.com/), or do we need this technic for any specific reason?

elia: That's not the case, although it might look like that at a first glance. For PCI compliance reasons the JS must be hosted on stripe.com and cannot be fully provided in the NPM package. In fact the package is nothing but a thin layer around loading the the JS from stripe.com.

The way it works is that if the `Stripe` constant is already available it will use that, and fallback to loading it on the fly if necessary, at the same time it provides the convenience of importing it as a module and having a Promise that can be waited on to ensure it's loaded.

elia: for the record: we discussed this live and removed the redundant global `<script>` tag in order to simplify the setup as the `stripe-js` import has a side-effect of adding it anyway as documented here: https://github.com/stripe/stripe-js#ensuring-stripejs-is-available-everywhere.

kennyadsl (approved):
👍

Issue #262, opened 2023-03-30
Detect refunds that have already been performed


waiting-for-dev:
On #180, we're handling the `charge.refunded` webhook event to generate a `Spree::Refund` copy. However, we're not checking whether the incoming webhook has already been processed in the past.

Although we're skipping the processing when the associated payment has already been fully refunded, it might be the case that we handle twice a partial amount refund. As noted in a [comment from the referenced issue](https://github.com/solidusio/solidus_stripe/issues/180#issuecomment-1484807184), we can't use the transaction ARN as that's something not always available at the moment the event is created.

There're two ways to handle that:

- Generally, through #188, making the system resilient against duplicated webhook events.
- Specifically, as another layer of security, associating every created refund with the charge id in another database table.

Another scenario for a duplicated refund would be one that has been initiated by Solidus, calling the Stripe API and getting the corresponding webhook event.

kennyadsl:
We can probably explore saving the id of the Stripe refund on the Solidus Refund in the `transaction_id` field. When the webhook comes, we could check if there's an existing refund and skip creating it on our end if so.

waiting-for-dev:
That could work. We'd need to save it both when we create a refund from Solidus admin and when we create it from an incoming webhook.

waiting-for-dev:
> We can probably explore saving the id of the Stripe refund on the Solidus Refund in the transaction_id field. 

The solution is trickier than expected. The problem is that Stripe doesn't add the created refund id on the event it sends when there's a refund (`charge.refunded`).

The best we can do is use the charge id to fetch all the refunds associated with it (`Stripe::Refund.list(charge: charge_id`) and create a refund on Solidus for all of the Stripe refunds that are not already present. That's to say, a `charge.refunded` event would handle all concurrent refunds: in the case of two partial refunds happening at the same time for a given order, the first `charge.refunded` event would trigger the creation of the two Solidus refunds, while the second one would do nothing.

Still, there's something else to consider. As we said in previous comments, a refund generated from Solidus will call the Stripe API and create a refund there, which in turn will publish the `charge.refunded` event. When we get the response from the Stripe API, we can update the Solidus refund `transaction_id` field with the Stripe refund id. However, there's no guarantee that, in the meantime, we haven't already processed the webhook (i.e., before updating the refund, therefore ending up with two records). We can workaround like this:
- Before calling the Stripe API, use a value `[[PENDING]]` for the created Solidus refund `transaction_id`.
[...]

PR #264, opened 2023-03-30
Handle `charge.refunded` event

### LINKED ISSUE #180: Handle `charge.refunded` webhook event
We'll get a `charge.refunded` webhook event whenever:

- A payment intent from a refundable payment method is refunded from Stripe's dashboard
- A Solidus payment is manually refunded from the admin interface (see #158)

We must update the Solidus counterpart for the first scenario by triggering the refund logic from the Solidus payment system. However, we need to be resilient because of the first scenario. We also need to ensure a refund is only performed once on Solidus, as the same webhook event could be received more than once. Because of that, we need to find a way to store the Acquirer Reference Number (ARN) assigned by Stripe to a given refund or, more generally, avoid any webhook being processed twice.

Partially replaces #160

waiting-for-dev:
Stripe [sends](https://stripe.com/docs/api/events/types#event_types-charge.refunded) a `charge.refunded` event when a refund is performed. When that happens, we need to accordingly update our data by creating a `Spree::Refund` for the given payment.

Notice that there are two ways of generating Solidus refunds: directly or through an RMA (Return Merchandise Authorization). A refund generated from Stripe can only integrate with the first kind, as there's no reliable way to hook into the second flow.

Stripe and Solidus support partial refunds. The event's `refund_amount` field we receive contains the sum of all the performed refunds on the payment, so we need to deduce the previously refunded amount before generating the new one.

Stripe refunds are required to be associated with a `Spree::RefundReason`, so we're seeding a record with a generic `Stripe refund` reason name. The name to fetch can be configured through the `#default_refund_reason_name` preference.

As noted in #262, we're not checking yet if an incoming refund has already been processed in the past.

Closes #180 

## Checklist


spec/lib/solidus_stripe/configuration_spec.rb
```diff
@@ -11,5 +11,11 @@
     it "defaults webhook signature tolerance to stripe's default" do
       expect(described_class.new.webhook_signature_tolerance).to eq(Stripe::Webhook::DEFAULT_TOLERANCE)
     end
+
+    it "defaults refund reason name to the one in seeds" do
+      expect(
+        described_class.new.default_refund_reason_name
+      ).to eq(SolidusStripe::Seeds::DEFAULT_STRIPE_REFUND_REASON_NAME)
+    end
```
kennyadsl, line 19: Not sure to follow who updates/creates a new refund reason when the preference is updated. Can you help me navigate this?

kennyadsl (commented):
Not sure to follow

spec/lib/solidus_stripe/configuration_spec.rb
waiting-for-dev: If another name is used, the developer must ensure a corresponding record exists in the database. Our seeds would provide a default, but they can manually change it if needed. Does it make sense?

lib/generators/solidus_stripe/install/templates/config/initializers/solidus_stripe.rb
```diff
@@ -11,6 +11,10 @@
   # Number of seconds while a webhook event is valid after its creation.
   # Defaults to the same value as Stripe's default.
   # config.webhook_signature_tolerance = 150
+  #
+  # Name of the `Spree::RefundReason` used for Stripe-generated refunds.
+  # Defaults to {SolidusStripe::DEFAULT_STRIPE_REFUND_REASON_NAME}.
+  # config.default_refund_reason_name = "Refund generated from Stripe"
```
kennyadsl: Shouldn't it be `refund_reason_name`? The default is this one: SolidusStripe::DEFAULT_STRIPE_REFUND_REASON_NAME, and when someone changes the preference value, they are no more using the default.

lib/generators/solidus_stripe/install/templates/config/initializers/solidus_stripe.rb
```diff
@@ -11,6 +11,10 @@
   # Number of seconds while a webhook event is valid after its creation.
   # Defaults to the same value as Stripe's default.
   # config.webhook_signature_tolerance = 150
+  #
+  # Name of the `Spree::RefundReason` used for Stripe-generated refunds.
```
kennyadsl, line 15: I'd add some lines to explain what to do when changing preference.

lib/solidus_stripe/seeds.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module SolidusStripe
+  # Seed data to seamlessly integrate Solidus with Stripe.
+  module Seeds
+    # Default name for the `Spree::RefundReason` used for Stripe refunds.
+    #
+    # Changed via {SolidusStripe::Configuration#default_refund_reason_name}.
+    #
+    # @return [String]
+    DEFAULT_STRIPE_REFUND_REASON_NAME = "Stripe Refund"
```
kennyadsl:
```suggestion
    DEFAULT_STRIPE_REFUND_REASON_NAME = "Refund generated from Stripe"
```

Personally, I prefer the reason text you used above, but not blocking.

kennyadsl (commented):
Ok, it's more clear now, thanks. Left another pass.

lib/generators/solidus_stripe/install/templates/config/initializers/solidus_stripe.rb
waiting-for-dev: :100: changed

lib/generators/solidus_stripe/install/templates/config/initializers/solidus_stripe.rb
waiting-for-dev: Done!

> If you change it, make sure that the corresponding `Spree::RefundReason` exists in the database with that name.

lib/solidus_stripe/seeds.rb
waiting-for-dev: :ok_hand: changed!

app/models/concerns/solidus_stripe/money_to_stripe_amount_converter.rb
```diff
@@ -72,6 +74,24 @@ def to_solidus_amount(fractional, currency)
     end
   end
 
+  # Transforms a decimal amount into a subunit amount
+  #
+  # @param amount [BigDecimal]
+  # @param currency [String]
+  # @return [Integer]
+  def solidus_decimal_to_subunit(amount, currency)
+    Money.from_amount(amount, currency).cents
```
elia: nit: Given the multiplier can vary and they're not always cents we could use `fractional` instead of `cents`, the two methods are aliased but `fractional` seems less misleading.

spec/models/concerns/solidus_stripe/money_to_stripe_amount_converter_spec.rb
```diff
@@ -7,44 +7,44 @@
   describe '.to_stripe_amount' do
     it 'converts to the fractional expected by stripe' do
       (described_class::ZERO_DECIMAL_CURRENCIES - %w[MGA]).each do |currency|
-        expect([currency, described_class.to_stripe_amount(12345, currency).to_i]).to eq([currency, 12345])
```
elia, line 10: I'm fine with the change, but just to give some context I think it was like that so that errors would include a reference to the currency when inside loops. If you think it's a worthy feature maybe it's better to add a comment for the next passerby wondering why it's done in this weird way 🤪

app/subscribers/solidus_stripe/webhook/charge_subscriber.rb
```diff
@@ -0,0 +1,77 @@
...
+      # Notice that, at this point, we have no way to distinguish between
+      # multiple occurrences of the same event.
+      #
+      # @param event [SolidusStripe::Webhook::Event]
+      def refund_payment(event)
+        event.data.object.to_hash => {
+          amount_refunded: new_stripe_total,
+          payment_intent: payment_intent_id,
+          currency:
+        }
```
elia, line 36: In case anyone else is wondering what this 👽 syntax is, it is the same as:

```rb
new_stripe_total = event.data.object.to_hash[:amount_refunded]
payment_intent = event.data.object.to_hash[:payment_intent_id]
currency = event.data.object.to_hash[:currency]
```

or, maybe better:

```rb
new_stripe_total, payment_intent, currency = event.data.object.to_hash.values_at(
  :amount_refunded, :payment_intent_id, :currency
)
```

app/subscribers/solidus_stripe/webhook/charge_subscriber.rb
```diff
@@ -0,0 +1,77 @@
...
+            message: "Payment was refunded after charge.refunded webhook (#{_1.money})"
+          )
+        end
+      end
+
+      private
+
+      def find_payment(payment_intent_id)
+        Spree::Payment.find_by!(response_code: payment_intent_id)
+      end
```
elia: nit: Thoughts on inlining to lower indirection while reading the code?

app/subscribers/solidus_stripe/webhook/charge_subscriber.rb
```diff
@@ -0,0 +1,77 @@
...
+        )
+      end
+
+      def refund_amount(new_stripe_total, currency, payment)
+        last_total = payment.refunds.sum(:amount)
+
+        new_stripe_total.
+          then { to_solidus_amount(_1, currency) }.
+          then { _1 - solidus_decimal_to_subunit(last_total, currency) }.
+          then { solidus_subunit_to_decimal(_1, currency) }
```
elia: nit: keeping the dot on the next line makes the git diff cleaner in case the last line is touched

lib/solidus_stripe/seeds.rb
```diff
@@ -0,0 +1,21 @@
...
+    #
+    # @return [String]
+    DEFAULT_STRIPE_REFUND_REASON_NAME = "Refund generated from Stripe"
+
+    def self.refund_reasons
+      Spree::RefundReason.find_or_initialize_by(
+        name: DEFAULT_STRIPE_REFUND_REASON_NAME
+      ).tap do
+        _1.update!(mutable: false) unless _1.persisted?
+      end
```
elia: I think here we can use `find_or_create_by` with a block, which is there precisely for the extra creation attributes:

```rb
Spree::RefundReason.find_or_initialize_by(name: DEFAULT_STRIPE_REFUND_REASON_NAME) do
  _1.mutable = false
end
```

https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-find_or_create_by

app/models/concerns/solidus_stripe/money_to_stripe_amount_converter.rb
waiting-for-dev: Makes a lot of sense. Thanks for the suggestion. Changed!

spec/models/concerns/solidus_stripe/money_to_stripe_amount_converter_spec.rb
waiting-for-dev: Oh, thanks for clarifying. I've added another commit using a custom message for the failure and adding the currency.

app/subscribers/solidus_stripe/webhook/charge_subscriber.rb
waiting-for-dev: It's probably that I tried to hide references to constants :slightly_smiling_face: Changed!

app/subscribers/solidus_stripe/webhook/charge_subscriber.rb
waiting-for-dev: Sure thing. Updated.

lib/solidus_stripe/seeds.rb
waiting-for-dev: Didn't know that :+1: Changed.

kennyadsl (approved):

rainerdema (approved):
Looks great! 💯

lib/solidus_stripe/configuration.rb
```diff
@@ -15,9 +16,15 @@ class Configuration
     #  its creation. Defaults to `Stripe::Webhook::DEFAULT_TOLERANCE`.
     attr_accessor :webhook_signature_tolerance
 
+    # @!attribute [rw] refund_reason_name
+    #   @return [String] `Spree::RefundReason#name` used for Stripe-generated
```
kennyadsl, line 20: Too much indent here?

elia (approved):

lib/solidus_stripe/configuration.rb
waiting-for-dev: I think it's ok. See https://rubydoc.info/gems/yard/file/docs/Tags.md#attribute

kennyadsl: Ah, because it refers to the attribute. Good to know, thanks!

PR #268, opened 2023-03-31
Handle partial captures

### LINKED ISSUE #221: Handle partial captures from the Stripe dashboard
## What

Stripe allows a Payment Intent to be partially captured. On the other side, Solidus only allows capturing 100% of payment.

We can detect that situation when receiving the `payment_intent.succeeded` webhook (see #179).

## Why

That difference in the domain model can lead to misalignment between the Stripe dashboard and the Store database. For instance, a refund on a partially captured payment could lead to different results if started from Stripe or Solidus.

## How

We must adapt because there's no way to disallow Stripe to perform partial captures. Different ideas:

- Warn when we detect that the captured payment doesn't match the total amount.
- Raise an error when we detect that the captured payment doesn't match the total amount.
- Update the Solidus payment amount to equal what's been captured. We need to check if something else needs to be done on the Solidus end.

waiting-for-dev:
## Summary

Stripe allows for a payment to be [partially captured](http://localhost:3000/checkout/confirm), creating a refund for the remaining amount.

We're now acting accordingly by changing how we handle the `payment_intent.succeeded` webhook. When the capture is total, we do as before. For partial captures, on top of the latter, we create a Solidus refund for the uncaptured amount.

Closes #221

## Checklist

### DIFF
--- a/app/subscribers/solidus_stripe/webhook/payment_intent_subscriber.rb
+++ b/app/subscribers/solidus_stripe/webhook/payment_intent_subscriber.rb
@@ -1,12 +1,15 @@
 # frozen_string_literal: true
 
+require "solidus_stripe/money_to_stripe_amount_converter"
+
 module SolidusStripe
   module Webhook
     # Handlers for Stripe payment_intent events.
     class PaymentIntentSubscriber
       include Omnes::Subscriber
+      include SolidusStripe::MoneyToStripeAmountConverter
 
-      handle :"stripe.payment_intent.succeeded", with: :complete_payment
+      handle :"stripe.payment_intent.succeeded", with: :capture_payment
       handle :"stripe.payment_intent.payment_failed", with: :fail_payment
       handle :"stripe.payment_intent.canceled", with: :void_payment
 
@@ -15,17 +18,26 @@ module SolidusStripe
       # Marks a Solidus payment associated to a Stripe payment intent as
       # completed, adding a log entry about the event.
       #
+      # In the case of a partial capture, a refund is created for the
+      # remaining amount and a log entry is added.
+      #
       # @param event [SolidusStripe::Webhook::Event]
-      def complete_payment(event)
+      def capture_payment(event)
         payment = extract_payment_from_event(event)
         return if payment.completed?
 
-        payment.complete!.tap do
-          SolidusStripe::LogEntries.payment_log(
-            payment,
-            success: true,
-            message: "Capture was successful after payment_intent.succeeded webhook"
-          )
+        event.data.object.to_hash => {
+          amount: stripe_amount,
+          amount_received: stripe_amount_received,
+          currency:
+        }
+        if stripe_amount == stripe_amount_received
+          complete_payment(payment)
+        else
+          payment.transaction do
+            complete_payment(payment)
+            refund_payment(payment, stripe_amount, stripe_amount_received, currency)
+          end
         end
       end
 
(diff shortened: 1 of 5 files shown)


elia (approved):

app/subscribers/solidus_stripe/webhook/payment_intent_subscriber.rb
```diff
@@ -74,6 +80,36 @@ def extract_payment_from_event(event)
...
+        payment.complete!.tap do
+          SolidusStripe::LogEntries.payment_log(
+            payment,
+            success: true,
+            message: <<~MSG
+              Partial capture (#{payment.money} of #{previous_money}) was successful after payment_intent.succeeded webhook. Payment amount was updated
+            MSG
+          )
+        end
+      end
```
rainerdema: I'm not 100% sure about this part 🤔 
The fact is that even if we have a partial capture, in this scenario, we also have a partial refund.
There is a `charge.captured` event just before the `payment_intent.succeeded` with the following params:
- "amount"
- "amount_captured"
- "amount_refunded"

But even considering the `amount_received` from `payment_intent.succeeded` (which I think is also right) my question is:
Would it be better if we create a refund using the `amount_refunded` value (or the `amount - amount_received`) rather than updating the original payment amount? What do you think?

kennyadsl (approved):
👍

app/subscribers/solidus_stripe/webhook/payment_intent_subscriber.rb
waiting-for-dev: Hmm. I see your point. The thing is that we don't get a `charge.refunded` webhook event, although Stripe does show a "Partial refund" label. At first sight, I don't see an associated refund record, though. I'll study it further. I'll put the PR as WIP in the meantime. Thanks for your feedback.

waiting-for-dev: @rainerdema, I confirm that even if the Stripe UI and the received webhooks differ, a Stripe Refund object is created under the hood for partial captures. Therefore, it makes total sense also to create if on our end. Thanks for pointing that out. I'm going to work on that straightaway.

app/subscribers/solidus_stripe/webhook/payment_intent_subscriber.rb
```diff
@@ -74,6 +86,38 @@ def extract_payment_from_event(event)
...
+          SolidusStripe::LogEntries.payment_log(
+            payment,
+            success: true,
+            message: "Capture was successful after payment_intent.succeeded webhook"
+          )
+        end
+      end
+
+      def refund_payment(payment, stripe_amount, stripe_amount_received, currency)
+        refunded_amount = decimal_amount(stripe_amount - stripe_amount_received, currency)
```
kennyadsl, line 101: Maybe a silly question: can `refunded_amount` be negative?

rainerdema (approved):
Thanks for looking into this @waiting-for-dev 🙌 Looks great!

app/subscribers/solidus_stripe/webhook/payment_intent_subscriber.rb
waiting-for-dev: I don't think so. `stripe_amount` is the total payment amount, while `stripe_amount_received` is the captured amount. Stripe won't let you capture more than the authorized amount.

kennyadsl (approved):

PR #271, opened 2023-04-03
Webhook development support and generic payment method slug

### LINKED ISSUE #220: Consider scoping all payment method routes to its slug
## What

On #213, we introduced a slug to differentiate webhook endpoints for different Stripe payment method instances. We can consider using it for the rest of the payment method requests (actions on the Stripe callback controller).

- [ ] share the slug with the intents controller
- [ ] listen for webhooks as part of `bin/dev`
- [ ] support an env variable for the slug similarly to how we setup the payment method in the initializer

## Why

That would be more consistent. Conversely, it could make sense only to use the slug when there's a compelling reason and use Rails' default `:id` param otherwise. See https://github.com/solidusio/solidus_stripe/pull/213#discussion_r1131411856 for the complete discussion.

elia:
## Summary

- both after confirmation and webhook urls are under the same scope
- bin/dev will now listen for the supported incoming webhooks
- fixes #220 

## Checklist


app/models/solidus_stripe/slug_entry.rb
```diff
@@ -14,24 +14,7 @@ module SolidusStripe
...
-    # Finds the payment method associated with the given slug.
-    #
-    # @param slug [String]
-    # @raise [ActiveRecord::RecordNotFound] if no payment method is found
-    # @return [SolidusStripe::PaymentMethod]
-    def self.payment_method(slug)
-      find_by!(slug: slug).payment_method
-    end
+  class SlugEntry < ::Spree::Base
+    belongs_to :payment_method, class_name: 'SolidusStripe::PaymentMethod'
```
rainerdema, line 18: Maybe I'm missing something, but I do not generally understand why we have a model + table SlugEntry/WebhookEndpoint and not a simple `slug` column in `spree_payment_methods`?

rainerdema (commented):
I left a comment, but overall looks great! 👏

app/models/solidus_stripe/slug_entry.rb
elia: @rainerdema it's only because the payment methods table uses STI and we don't want to pollute it for every other payment method. Alternatively we could have used preferences but those can't be used in queries as they're serialized as YAML.

rainerdema: Makes total sense 🙌
> Alternatively we could have used preferences but those can't be used in queries as they're serialized as YAML.

Thinking about preferences, I think it can still be a good compromise given their purpose 🤔 
However, these changes fully represent the purpose of the related issue, and maybe we can discuss it in a separate PR.

Thanks @elia 🙏

rainerdema (approved):
💯

lib/solidus_stripe/webhook/event.rb
```diff
@@ -24,7 +24,7 @@ class Event
       # @api private
       class << self
         def from_request(payload:, signature_header:, slug:, tolerance: default_tolerance)
-          payment_method = SolidusStripe::WebhookEndpoint.payment_method(slug)
+          payment_method = SolidusStripe::PaymentMethod.with_slug(slug).first!
```
waiting-for-dev, line 27: Not sure to get why we unwrapped the procedure to get the payment method. IMHO, that opens gates to duplication and copy-paste programming.

app/models/solidus_stripe/payment_method.rb
```diff
@@ -70,12 +67,12 @@ def stripe_dashboard_url(intent_id)
       end
     end
 
-    private
+    def assign_slug
```
waiting-for-dev, line 70: I think responsibilities were fairer when that method was part of the `SlugEntry` class, especially considering that the payment method already deals with a lot of stuff (via inheritance from upstream Solidus).

app/models/solidus_stripe/payment_method.rb
```diff
@@ -70,12 +67,12 @@ def stripe_dashboard_url(intent_id)
...
+      # If there's only one payment method, we can use a default slug.
+      slug = preferred_test_mode ? 'test' : 'live' if self.class.count == 1
+      slug = SecureRandom.hex(16) while SlugEntry.exists?(slug: slug) || slug.nil?
```
waiting-for-dev, line 73: For sure opinionated, but IMHO using recursion made it simpler (for instance, we didn't need to check for `slug.nil?`) :slightly_smiling_face:

waiting-for-dev (commented):
Thanks for working on that.

I left a posthumous review. No need to take action, but leaving my thoughts for future reference.

lib/solidus_stripe/webhook/event.rb
elia: @waiting-for-dev that's a valid concern. The reason it was moved to a slightly lower-level construct like a scope is that it's more versatile and can be mixed and matched with other scopes without generating extra queries more easily, e.g. in [this case](https://github.com/solidusio/solidus_stripe/pull/271/files/731629d88e5e2dea570686dd412a45b0fd77cb30#diff-731d975f879e89e2da5e248ccdfdcf53d9aebf9a8d5add2289218a32f23b2627R47)

https://github.com/solidusio/solidus_stripe/blob/731629d88e5e2dea570686dd412a45b0fd77cb30/app/controllers/solidus_stripe/intents_controller.rb#L45-L48

app/models/solidus_stripe/payment_method.rb
elia: I think it was originally moved in order to check the `.count` and `.preferred_test_mode` on the payment method, but could be simplified for sure, e.g. the following is probably equivalent:

```rb
  return if slug_entry.present?

  slug = preferred_test_mode ? 'test' : 'live'
  slug = SecureRandom.hex(16) while SlugEntry.exists?(slug: slug)
```

Although given how rare it is I agree we should do it for readability (vs. performance).

app/models/solidus_stripe/payment_method.rb
elia: The main reason was (I think) having access to `preferred_test_mode` (see the other comment), personally I'm fine with both arrangements, with a preference for SlugEntry not needing to know too much about the payment method.

elia (commented):
@waiting-for-dev I hope I provided some context on the code changes you highlighted, please feel free to send PRs for refining rough edges if there's anything you particularly care about 🙌

PR #278, opened 2023-04-13
Allow reuse of payment sources via admin

### LINKED ISSUE #267: Consider adding placeholder #authorize and #purchase methods to the gateway
## Why

In case anyone would read the gateway class implementations we should avoid the surprise of not finding those two methods, that are mentioned in the solidus guides for creating custom gateways.

## What

Add the methods along with documentation, consider using either a `raise NotImplementedError` or a `# noop` comment as the implementation.

This might be impacted by #266 in case those methods will be implemented because of it.

### LINKED ISSUE #280: Allow admin to reuse existing payment sources
## Introduction

The goal of this issue is to allow the creation of admin-side payments using the Solidus Stripe payment method, by reusing payment sources through the admin interface.

## Current Behavior

Currently, admins can only reuse existing payment sources via the frontend and not the backend.

## Desired Behavior

To allow the admin to reuse existing payment sources, a [source form](https://guides.solidus.io/advanced-solidus/payments-and-refunds/#providing-payment-method-partials) needs to be implemented for the Solidus Stripe Payment Method.
This would enable the reuse of payment sources within the user's wallet, preventing the use of payment methods that the user has removed from their wallet.

## Additional Considerations

Since Stripe offers various payment methods beyond credit cards, partial payment-method-type views should be dedicated to each Stripe Payment Method to render the different parameters related to that specific type:
[Something similar to what we're already doing for the frontend-side](https://github.com/solidusio/solidus_stripe#showing-reusable-sources-in-the-checkout)

## Possible Problems

The current customization in Solidus does not fully support payment sources that differ from credit cards (in terms of methods and attributes nomenclature).
In the specific case of Stripe, this is incorrect since a Stripe payment method (or a payment source in Solidus) can represent many types of payments other than credit cards.

You can analyze this problem in:
- [The sources passed inside the source form, which is called `previous_cards`](https://github.com/solidusio/solidus/blob/e9debb976e2228bb0b7a8eff4894e0556fc15cc8/backend/app/views/spree/admin/payments/_form.html.erb#L31)
- [The expected attribute in the controller called `card`](https://github.com/solidusio/solidus/blob/e9debb976e2228bb0b7a8eff4894e0556fc15cc8/backend/app/controllers/spree/admin/payments_controller.rb#L28)

For now, we should adhere to this nomenclature, even though it is not customizable.

## TODO

- [x] We need to open an issue regarding the mentioned problems on Solidus: https://github.com/solidusio/solidus/issues/5014

rainerdema:
## Summary

- Fixes #280
- Fixes #267 

ℹ️  Please refer to #280 for more details about specific implementation decisions.

- [x] Allow the admin to reuse existing payment sources by introducing a dedicated [source form](https://guides.solidus.io/advanced-solidus/payments-and-refunds/#providing-payment-method-partials). 
  This enables the reuse of payment sources within the user's wallet (preventing the use of payment methods that the user has removed from their wallet).
- [x] Re-introduce `authorize` and `purchase` gateway methods

Screenshot 2023-04-13 at 20 03 09

## Checklist

### DIFF
--- a/app/models/solidus_stripe/gateway.rb
+++ b/app/models/solidus_stripe/gateway.rb
@@ -31,6 +31,44 @@ module SolidusStripe
 
     attr_reader :client
 
+    # Authorizes a certain amount on the provided payment source.
+    #
+    # We create and confirm the Stripe payment intent in two steps. That's to
+    # guarantee that we associate a Solidus payment on the creation, and we can
+    # fetch it after the webhook event is published on confirmation.
+    #
+    # The Stripe payment intent id is copied over the Solidus payment
+    # `response_code` field.
+    def authorize(amount_in_cents, _source, options = {})
+      check_given_amount_matches_payment_intent(amount_in_cents, options)
+
+      stripe_payment_intent_options = {
+        amount: to_stripe_amount(amount_in_cents, options[:originator].currency),
+        confirm: false,
+        capture_method: "manual",
+      }
+
+      stripe_payment_intent = SolidusStripe::PaymentIntent.prepare_for_payment(
+        options[:originator],
+        **stripe_payment_intent_options,
+      )
+
+      confirm_stripe_payment_intent(stripe_payment_intent.stripe_intent_id)
+
+      build_payment_log(
+        success: true,
+        message: "PaymentIntent was confirmed successfully",
+        response_code: stripe_payment_intent.stripe_intent_id,
+        data: stripe_payment_intent.stripe_intent
+      )
+    rescue Stripe::StripeError => e
+      build_payment_log(
+        success: false,
+        message: e.message,
+        data: e.response
+      )
+    end
+
     # Captures a certain amount from a previously authorized transaction.
     #
     # @see https://stripe.com/docs/api/payment_intents/capture#capture_payment_intent
@@ -49,6 +87,41 @@ module SolidusStripe
         data: payment_intent,
       )
     rescue Stripe::InvalidRequestError => e
+      build_payment_log(
+        success: false,
+        message: e.to_s,
+        data: e.response,
+      )
+    end
+
+    # Authorizes and captures a certain amount on the provided payment source.
+    #
+    # See `#authorize` for how the confirmation step is performed.
+    #
+    # @todo add support for purchasing custom amounts
+    def purchase(amount_in_cents, _source, options = {})
+      check_given_amount_matches_payment_intent(amount_in_cents, options)
+
+      stripe_payment_intent_options = {
+        amount: to_stripe_amount(amount_in_cents, options[:originator].currency),
+        confirm: false,
+        capture_method: "automatic",
+      }
+
+      stripe_payment_intent = SolidusStripe::PaymentIntent.prepare_for_payment(
+        options[:originator],
+        **stripe_payment_intent_options,
+      )
+
+      confirm_stripe_payment_intent(stripe_payment_intent.stripe_intent_id)
+
+      build_payment_log(
+        success: true,
+        message: "PaymentIntent was confirmed and captured successfully",
+        response_code: stripe_payment_intent.stripe_intent_id,
+        data: stripe_payment_intent.stripe_intent,
+      )
+    rescue Stripe::StripeError => e
       build_payment_log(
         success: false,
         message: e.to_s,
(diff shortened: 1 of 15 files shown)


app/models/solidus_stripe/gateway.rb
```diff
@@ -56,6 +94,35 @@ def capture(amount_in_cents, payment_intent_id, options = {})
...
+      )
+
+      confirm_stripe_payment_intent(stripe_payment_intent.stripe_intent_id)
+
+      build_payment_log(
+        success: true,
+        message: "PaymentIntent was confirmed and captured successfully",
+        response_code: stripe_payment_intent.id,
+        data: stripe_payment_intent.stripe_intent,
+      )
```
elia, line 123: Should we add error handling here as well?

app/models/solidus_stripe/payment_intent.rb
```diff
@@ -23,6 +23,15 @@ def self.prepare_for_payment(payment)
       intent
     end
 
+    def self.retrieve_last_usable_intent(payment)
+      intent = where(payment_method: payment.payment_method, order: payment.order).last
+      intent if intent&.usable?
+    end
+
+    def usable?
+      stripe_intent_id && stripe_intent.status == 'requires_payment_method'
```
elia: Thoughts on checking the amount too?

elia (commented):
Just a couple of nits

💯 on having an issue open on solidus, we should mark it for fixing whenever we touch that part of the admin (at worst).

app/models/solidus_stripe/payment_intent.rb
rainerdema: Yes! You're right. This could generate dangerous problems. Thanks @elia I'll verify the amount ⏳

app/models/solidus_stripe/gateway.rb
rainerdema: 🙏 ✔️ 
https://github.com/solidusio/solidus_stripe/blob/140bfbcb23c0bb8a4a693d9cb6b050056cd19958/app/models/solidus_stripe/gateway.rb#L124-L129

app/models/solidus_stripe/gateway.rb
```diff
@@ -56,6 +94,41 @@ def capture(amount_in_cents, payment_intent_id, options = {})
...
+        confirm: false,
+        capture_method: "automatic",
+      }
+
+      stripe_payment_intent = SolidusStripe::PaymentIntent.prepare_for_payment(
+        options[:originator],
+        **stripe_payment_intent_options,
+      )
+
+      confirm_stripe_payment_intent(stripe_payment_intent.stripe_intent_id)
```
waiting-for-dev, line 116: I think we're missing the capture part, right? :thinking:

app/models/solidus_stripe/payment_method.rb
```diff
@@ -82,5 +82,25 @@ def assign_slug
...
+    # https://github.com/solidusio/solidus/blob/e9debb976e2228bb0b7a8eff4894e0556fc15cc8/backend/app/views/spree/admin/payments/_form.html.erb#L31
+    # This name is inaccurate and too specific because, in our case, a
+    # payment-source/stripe-payment-method have many different possible types:
+    # https://stripe.com/docs/api/payment_methods/object#payment_method_object-type
+    #
+    # For more details:
+    # https://github.com/solidusio/solidus/issues/5014
+    #
+    # @todo Start using the correct method to get a user's previous sources
+    def previous_sources(order)
```
waiting-for-dev, line 98: Should we add tests for this method?

app/models/solidus_stripe/gateway.rb
waiting-for-dev: Should we add tests?

waiting-for-dev (commented):
Thanks, @rainerdema! Great work and great investigation :clap: Left some comments.

app/models/solidus_stripe/gateway.rb
rainerdema: I don't think so, as the mentioned `purchase` action has an `automatic` capture_mode setting.
The capture should be included in the confirmation process 🤔

waiting-for-dev: Oh, yeah, you're right. Sorry for the noise

waiting-for-dev (approved):
:raised_hands:

PR #281, opened 2023-04-14
Avoid duplicated Solidus refunds

### LINKED ISSUE #262: Detect refunds that have already been performed

waiting-for-dev:
## Summary

Stripe doesn't guarantee that a webhook event will be delivered only once, so we need to be prepared to handle duplicated incoming refund events. Just creating a Solidus copy whenever we receive a refund event is not enough.

On top of that, Stripe won't identify the refund that triggered the event. Both the `charge.refunded` and `payment_intent.succeeded` (for partial captures, which generates a refund for the remaining amount under the hood) webhooks only give us information to retrieve the list of all refunds associated with a charge or payment intent.

Because of this, we are now syncing all the refunds associated with a payment intent at every webhook event. We keep track of the refunds already present on Solidus by leveraging the `transaction_id` field on `Spree::Refund`, copying the Stripe refund id as value.

On top of that, we need to account for refunds created from Solidus admin panel, which calls the Stripe API. In this case, we need to avoid syncing the refund back to Solidus on the subsequent webhook. We're marking those refunds with a metadata field on Stripe, so we can exclude them from the sync.

Closes #262

## Checklist

### DIFF
--- a/app/models/solidus_stripe/gateway.rb
+++ b/app/models/solidus_stripe/gateway.rb
@@ -2,6 +2,7 @@
 
 require 'stripe'
 require "solidus_stripe/money_to_stripe_amount_converter"
+require "solidus_stripe/refunds_synchronizer"
 
 module SolidusStripe
   # @see https://stripe.com/docs/payments/accept-a-payment?platform=web&ui=checkout#auth-and-capture
@@ -78,6 +79,11 @@ module SolidusStripe
     end
 
     # Refunds the provided amount on a previously captured transaction.
+    #
+    # Notice we're adding `solidus_skip_sync: 'true'` to the metadata to avoid a
+    # duplicated refund after the generated webhook event. See
+    # {RefundsSynchronizer}.
+    #
     # TODO: check this method params twice.
     def credit(amount_in_cents, payment_intent_id, options = {})
       check_payment_intent_id(payment_intent_id)
@@ -89,13 +95,16 @@ module SolidusStripe
         Stripe::Refund.create(
           amount: to_stripe_amount(amount_in_cents, currency),
           payment_intent: payment_intent_id,
+          metadata: {
+            RefundsSynchronizer::SKIP_SYNC_METADATA_KEY => RefundsSynchronizer::SKIP_SYNC_METADATA_VALUE
+          }
         )
       end
 
       build_payment_log(
         success: true,
         message: "PaymentIntent was refunded successfully",
-        response_code: payment_intent_id,
+        response_code: stripe_refund.id,
         data: stripe_refund,
       )
     end
--- a/lib/solidus_stripe/refunds_synchronizer.rb
+++ b/lib/solidus_stripe/refunds_synchronizer.rb
@@ -1,0 +1,96 @@
+# frozen_string_literal: true
+
+require "solidus_stripe/money_to_stripe_amount_converter"
+
+module SolidusStripe
+  # Synchronizes refunds from Stripe to Solidus.
+  #
+  # For our use case, Stripe has two ways to inform us about refunds initiated
+  # on their side:
+  #
+  # 1. The `charge.refunded` webhook event, which is triggered when a refund is
+  # explicitly created.
+  # 2. The `payment_intent.succeeded` webhook event, which is triggered when a
+  # payment intent is captured. If the payment intent is captured for less than
+  # the full amount, a refund is automatically created for the remaining amount.
+  #
+  # In both cases, Stripe doesn't tell us which refund was recently created, so
+  # we need to fetch all the refunds for the payment intent and check if any of
+  # them is missing on Solidus. We're using the `transaction_id` field on
+  # `Spree::Refund` to match refunds against Stripe refunds ids. We could think
+  # about only syncing the single refund not present on Solidus, but we need to
+  # acknowledge concurrent partial refunds.
+  #
+  # The `Spree::RefundReason` with `SolidusStripe::Config.refund_reason_name`
+  # as name is used as created refunds' reason.
+  #
+  # Besides, we need to account for refunds created from Solidus admin panel,
+  # which calls the Stripe API. In this case, we need to avoid syncing the
+  # refund back to Solidus on the subsequent webhook, otherwise we would end up
+  # with duplicate records. We're marking those refunds with a metadata field on
+  # Stripe, so we can filter them out (see {Gateway#credit}).
+  class RefundsSynchronizer
+    include MoneyToStripeAmountConverter
+
+    # Metadata key used to mark refunds that shouldn't be synced back to Solidus.
+    # @return [Symbol]
+    SKIP_SYNC_METADATA_KEY = :solidus_skip_sync
+
+    # Metadata value used to mark refunds that shouldn't be synced back to Solidus.
+    # @return [String]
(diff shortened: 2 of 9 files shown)


lib/solidus_stripe/refunds_synchronizer.rb
```diff
@@ -0,0 +1,94 @@
...
+    private
+
+    def stripe_refunds(payment_intent_id)
+      @payment_method.gateway.request do
+        Stripe::Refund.list(payment_intent: payment_intent_id).data
+      end
+    end
+
+    def stripe_refund_needs_sync?(refund)
+      (refund.metadata[SKIP_SYNC_METADATA_KEY] != SKIP_SYNC_METADATA_VALUE) &&
```
kennyadsl: Not blocking, especially because this class is well documented, but these two lines requires some knowledge to understand what happens here (which I consider the most important logic of the class). I would personally move them to another method or variable that express what they do, something like:


```
def stripe_refund_needs_sync?(refund)
  originated_outside_solidus = refund.metadata[SKIP_SYNC_METADATA_KEY] != SKIP_SYNC_METADATA_VALUE)
  not_yet_persisted_in_solidus = Spree::Refund.find_by(transaction_id: refund.id).nil?
  
  originated_outside_solidus && not_yet_persisted_in_solidus
end
     
```

Again, I'm perfectly fine with what we have, and I know this version is too oriented for future readers' happiness, but I think it was worth sharing.

kennyadsl (approved):
👍 Left a comment, but I'm fine with this, thanks @waiting-for-dev

lib/solidus_stripe/refunds_synchronizer.rb
waiting-for-dev: I think it makes total sense. Changed!!

lib/solidus_stripe/refunds_synchronizer.rb
```diff
@@ -0,0 +1,96 @@
...
+    SKIP_SYNC_METADATA_VALUE = 'true'
+
+    # @param payment_method [SolidusStripe::PaymentMethod]
+    def initialize(payment_method)
+      @payment_method = payment_method
+    end
+
+    # @param payment_intent_id [String]
+    def call(payment_intent_id)
+      payment = Spree::Payment.find_by!(response_code: payment_intent_id)
```
elia, line 50: Thoughts on restricting to the payment method? `@payment_method.payments.find_by(…)`

app/models/solidus_stripe/gateway.rb
```diff
@@ -89,13 +95,16 @@ def credit(amount_in_cents, payment_intent_id, options = {})
         Stripe::Refund.create(
           amount: to_stripe_amount(amount_in_cents, currency),
           payment_intent: payment_intent_id,
+          metadata: {
+            RefundsSynchronizer::SKIP_SYNC_METADATA_KEY => RefundsSynchronizer::SKIP_SYNC_METADATA_VALUE
+          }
         )
       end
```
elia, line 102: Thoughts on letting the synchronizer do this and avoid coupling the metadata key/value?
In other words: "if you go through the synchronizer we can guarantee you won't get duplicates"

lib/solidus_stripe/refunds_synchronizer.rb
```diff
@@ -0,0 +1,96 @@
...
+
+      stripe_refunds(payment_intent_id)
+        .select(&method(:stripe_refund_needs_sync?))
+        .map(
+          &method(:create_refund).curry[payment]
+        )
```
elia, line 56: Thoughts on just using a block here `.map { create_refund(payment, _1) }` for the sake of readability?

lib/solidus_stripe/refunds_synchronizer.rb
```diff
@@ -0,0 +1,96 @@
...
+      originated_outside_solidus && not_already_synced
+    end
+
+    def create_refund(payment, refund)
+      Spree::Refund.create!(
+        payment: payment,
+        amount: refund_decimal_amount(refund),
+        transaction_id: refund.id,
+        reason: SolidusStripe::PaymentMethod.refund_reason
+      ).tap(&method(:log_refund).curry[payment])
```
elia, line 80: Also here, maybe `.tap { log_refund(payment, _1) }` would be more readable.

elia:
@waiting-for-dev sorry if the review took long, maybe some of the comments can be addressed in a follow up if they make sense to you

app/models/solidus_stripe/gateway.rb
waiting-for-dev: Not sure to understand. The synchronizer takes refunds from Stripe and copies them to Solidus if not present, while the gateway is responsible for creating stuff from Solidus to Stripe. :thinking:

elia: Both SolidusStripe::PaymentIntent and SolidusStripe::Customer take charge of a specific part of the Stripe API, including creation. My suggestion is that maybe that should be the case for the RefundSynchronizer as well. Considering a smell this cross reference of metadata.

A less invasive alternative would be to simplify this part doing something like:

```rb
metadata: {
  **RefundsSynchronizer.stripe_metadata
}
```

lib/solidus_stripe/refunds_synchronizer.rb
waiting-for-dev: An id collision is probably impossible, and the query will be slightly slower, but that won't hurt.

app/models/solidus_stripe/gateway.rb
waiting-for-dev: TBH, I think having API calls spread across the codebase is a bigger smell :slightly_smiling_face: Ideally, I'd like them to get closer to the gateway, wrapped within an adapter class, if that would be too much for it.

So I'll go with your latest suggestion, which I think best encapsulates details.

lib/solidus_stripe/refunds_synchronizer.rb
waiting-for-dev: :+1: I understand it's not the most standard Ruby, although when you get used to it readability is even better. What I would love is Ruby to support something like:

```ruby
.map(&create_refund(payment))
```

lib/solidus_stripe/refunds_synchronizer.rb
waiting-for-dev: :+1:

waiting-for-dev:
Thanks for your review @elia. Please, see https://github.com/solidusio/solidus_stripe/pull/290

PR #284, opened 2023-04-17
Stop depending on solidus_dev_support

elia:
## Summary

With the new way of testing solidus extensions (through full rails installations) we no longer have much to gain from using solidus_dev_support. 

Specifically the only two things we were using it for were:
1. rubocop rules
2. releasing the gem

### Rubocop rules

We just had a few dev-support adjustments over the "relaxed" rubocop style, but we also had a handful of local rules.

It seems ok to just import the original list and move from there, all modern tooling will adapt to whatever rules are in the repo (language servers, formatters, etc.).

### Releasing

The dev-support releasing toolchain starter at a time in which GitHub releases were not so convenient (if they existed at all). Now it's super easy to store the changelog there directly and be generated with one click.

This of course is ok for extensions as the have a smaller scope and simpler changes (vs. Solidus).

## Checklist


rainerdema (approved):
👍

README.md
```diff
@@ -235,7 +235,12 @@ Use Ctrl-C to stop
 
 ### Releasing new versions
 
-Please refer to the dedicated [page](https://github.com/solidusio/solidus/wiki/How-to-release-extensions) on Solidus wiki.
+1. Ensure that the CI is green
+2. Update the version in `lib/solidus_stripe/version.rb`
+3. Commit the changes with the message "Release vX.Y.Z"
+4. Push the commit
+5. Run `bin/rake release`
+6. Open https://github.com/solidusio/solidus_stripe/releases/vX.Y.Z and generate the changelog
```
waiting-for-dev: I like this flow more than the previous one. However, I'm afraid of having specific instructions for this extensions. What do you think about moving the steps to the previously referenced wiki page on Solidus and making them the recommended way to release extensions? In the end, there's no reason why extensions still using `solidus_dev_support` can't use them. Then, when we find time, we can remove the changelog generator from `solidus_dev_support`. Thoughts? cc @kennyadsl, @rainerdema

waiting-for-dev (changes requested):

README.md
kennyadsl: Yes, I lately use that wiki page only when releasing extensions. BTW, I'm more for starting to move shared things back to `dev_support`, so that all the extensions can benefit from the progress made here and on the other payment gateway extensions. 

In other words, I'm scared to go back to a situation where we need to manually update all the extensions when something happens (change to the sandbox script, or for a dependency constraint).

elia: I'm fine with moving stuff back to dev-support, but I thought it wasn't the time yet, and I think we need to think through how we want to handle legacy extensions with regard to that. 

That's why I kept everything local for stripe, so it can be the place from which we can extract the whole recipe, maybe in a new major of dev-support that kinda does things from scratch.

This would probably include sharing some of these scripts and templates with the starter frontend, which poses even more questions on how to deal with things, not being a gem itself.

waiting-for-dev: I agree we should plan when the time come to move stuff to `solidus_dev_support`. But what about start using the new release process now? That doesn't require any change on `solidus_dev_support`. What I propose is just moving these instructions to the solidus wiki.

kennyadsl: Yes maybe it's not time yet, but can't what we described here for releasing be applied to all extensions by just changing the wiki page?

elia: I opened https://github.com/solidusio/solidus_dev_support/issues/202 to track backporting things to dev-support.

This extension is the ultimate blueprint for developing solidus extensions for the starter frontend, and whatever we do here is likely to become the standard and be codified in dev-support, eventually. I also think it's better not to block solidus_stripe on changes to dev-support to avoid both further delays and rushing decisions.

waiting-for-dev: @elia, I think we're not being able to communicate our intentions properly :sweat_smile:

> it's better not to block solidus_stripe on changes to dev-support to avoid both further delays and rushing decisions

Absolutely, we're not advocating for that! Our proposal is as follows:

- Copying the following:
```
1. Ensure that the CI is green
2. Update the version in `lib/solidus_stripe/version.rb`
3. Commit the changes with the message "Release vX.Y.Z"
4. Push the commit
5. Run `bin/rake release`
6. Open https://github.com/solidusio/solidus_stripe/releases/vX.Y.Z and generate the changelog
```
- Pasting it into https://github.com/solidusio/solidus/wiki/How-to-release-extensions (replacing `solidus_stripe` by something like `your_extension`).
- Removing the old content in https://github.com/solidusio/solidus/wiki/How-to-release-extensions, maybe adding a line about installing `gem-release` if not present.
- Remove the instructions from this PR and just have a link to the Solidus wiki.

Your new proposal is better. As you said, with GH releases and the automatic release notes generator, everything is a piece of cake. There's no need to use `solidus_dev_support` for any extension. And we want to avoid having instructions that can differ from one extension to another.

I hope this is clear now :slightly_smiling_face: What do you think about it?

elia: @waiting-for-dev I was under the impression there were multiple positions on this, thanks for clarifying 🙌 

The wiki entry is up to date, I also added instructions on how to migrate older extensions and kept the older instructions under a `<summary>` in case there's no time to migrate an extension.

https://github.com/solidusio/solidus/wiki/How-to-release-extensions

waiting-for-dev (approved):

waiting-for-dev (changes requested):
Sorry @elia, we still need to remove the instructions on the README so they don't get staled. The wiki should be the unique source of truth.

elia:
@waiting-for-dev done ✅

waiting-for-dev (approved):
Lovely!

PR #293, opened 2023-04-21
update doc after using the slug as a general scope

loicginoux:
the change introduced in commit [731629d88e5e2dea570686dd412a45b0fd77cb30](https://github.com/solidusio/solidus_stripe/commit/731629d88e5e2dea570686dd412a45b0fd77cb30) forgot to mention how the url changed in doc


elia:
Thanks, we're in the process of reviewing the README, I'll incorporate your commit in that PR 👍 🙏

Issue #208, opened 2023-03-07
QA the documentation


elia:
## Why

We need to ensure all options and tradeoffs are clear to users, as well as the limitations of the default setup.

## What

This includes:
~- payment/setup intents difference~
~- skipping the confirm step~

- [x] authorizing at the ~payment~ confirm step
- [x] implications for non card methods and Europe
- [x] how to migrate to the new extension
- [x] how to integrate with a different frontend

chrean:
We can make the README as good as possible to keep documentation as close to the code as possible.
Then we may add a page in the guides with some basic information and a link to the README.

PR #295, opened 2023-04-21
Allow the use of non-card Stripe payment methods

### LINKED ISSUE #175: Submit billing information to Stripe
We probably want to push the billing information to Stripe. At this point, we're only sending the user email.

### LINKED ISSUE #263: Manage Payment Method Types available
**Introduction**

Customers see the available payment methods during the checkout process.
We can either manage payment methods from the [Stripe Dashboard](https://dashboard.stripe.com/settings/payment_methods) or list payment methods manually in code.

**Current Behavior**

Currently, there is no specific setting, and this, by default, means that all payment methods enabled on Stripe are rendered in our checkout process.

**Possible problems related to the current implementation**
- For each payment method, we need a [partial view dedicated to rendering the saved payment sources](https://github.com/solidusio/solidus_stripe/blob/59f440383c7d86dc195d4090a16c4386185eba22/lib/generators/solidus_stripe/install/templates/app/views/orders/payment_info/_stripe.html.erb#L2-L5) (see this example for [credit cards](https://github.com/solidusio/solidus_stripe/blob/master/lib/generators/solidus_stripe/install/templates/app/views/checkouts/existing_payment/stripe/_card.html.erb)). This means that although a new payment method is enabled on Stripe, this also requires a change in the Solidus store, which may make it necessary to decide not to enable the payment method automatically.

- Another problem is related to the simple management of the store. The same Stripe account can be used in multiple environments. For each of these, it is necessary to give the possibility to decide which payment methods to enable regardless of the global setting on the Stripe Dashboard.

**Desired Behavior**

We need to improve the management of available payment methods to address the possible scenarios mentioned.
This involves updating the Stripe Elements creation and the correlated Payment Intent creation.

There are two ways we can achieve this:
- Use the [`automatic_payment_methods`](https://stripe.com/docs/api/payment_intents/object#payment_intent_object-automatic_payment_methods-enabled), which will automatically include any payment methods activated on the Stripe dashboard.
- Specify the payment methods we want to make available by including a [`payment_method_types`](https://stripe.com/docs/api/payment_intents/object#payment_intent_object-payment_method_types) parameter.

We can implement one or both of these approaches.

**Additional notes**

Stripe Elements:
- [`payment_method_types`](https://stripe.com/docs/js/elements_object/create_without_intent#stripe_elements_no_intent-options-paymentMethodTypes)

Stripe Payment Intent:
- [`automatic_payment_methods`](https://stripe.com/docs/api/payment_intents/object#payment_intent_object-automatic_payment_methods-enabled)
- [`payment_method_types`](https://stripe.com/docs/api/payment_intents/object#payment_intent_object-payment_method_types)

It should be noted that having the basic automatic setting is recommended. As specified in the [Stripe documentation](https://stripe.com/docs/payments/payment-methods/integration-options#using-automatic-payment-methods):
>The payments acceptance page prioritizes showing payment methods known to increase conversion for your customer’s location while lower priority payment methods are hidden in an overflow menu.

This means that having a manual set of payment method types could penalize this behavior: This should be specified in the [README.md](https://github.com/solidusio/solidus_stripe/blob/master/README.md).

For more information, please refer to:
- [Choose how to add payment methods](https://stripe.com/docs/payments/payment-methods/integration-options#choose-how-to-add-payment-methods)
- [Using automatic payment methods with the Payment Element](https://youtu.be/5lS6WdT2GUI?t=384)

### LINKED ISSUE #300: Problems detecting and using Stripe payment methods other than credit cards
**Introduction**

Even if there is already partial support for non-card Stripe Payment Methods, they still need to be fully supported.

**Current behavior**

When trying to add a `sepa_debit` payment method type, by enabling it from the Stripe dashboard, these are the problems encountered:
- Stripe Payment Methods cannot be attached to the created Stripe Payment Intent if this doesn't have the `automatic_payment_methods` enabled.
- The [default payment method partial view (dedicated to existing payment methods and details in the confirm step)](https://github.com/solidusio/solidus_stripe/blob/aa50d74f08619b1d60ff5163f63d9c0ec14a0864/lib/generators/solidus_stripe/install/templates/app/views/checkouts/existing_payment/stripe/_default.html.erb#L1-L4) is not rendered correctly and generates an exception due to missing params.
- Stripe Payment Intent "processing" status, allowed in some payment methods like "sepa_debit", [is not managed, so the payment is marked as failed in Solidus](https://github.com/solidusio/solidus_stripe/blob/aa50d74f08619b1d60ff5163f63d9c0ec14a0864/app/models/solidus_stripe/payment_intent.rb#L52).

**Expected behavior**

It's expected to see the supported and enabled Stripe payment methods in the checkout payment section and be able to use them (even without a dedicated configuration other than the one made for credit cards).

rainerdema:
## Summary

- Fixes #300
- Fixes #263
- Documentation for this is already covered by #208
- Fixes #175 

The goal of these changes is to see the supported and enabled Stripe payment methods in the checkout payment section and be able to use them (even without a dedicated configuration like dedicated partial views for payment method types).

Please look at the attached issues and all the commit details for more context.

### Testing

https://user-images.githubusercontent.com/19948291/234878333-9c177079-a25f-4f57-850f-7f71700ac5f3.mov

In case of manual tests, be sure to:
1. _Activate the payment methods_ you want to test from the [Stripe Dashboard](https://dashboard.stripe.com/test/settings/payment_methods).
2. Set the `Auto Capture` mode to `Yes` in your Solidus Stripe Payment Method:
This is needed to allow the display of non-card payment methods that don't support `capture_mode` set to `false`: Most of the Stripe payment methods have this behavior.
3. _Have the correct currency_ based on your Stripe Country account: For example, if my Stripe account is configured in "IT" I will probably need an order with "EUR" currency to display some of the activated payment methods.
4. Try to use a _Stripe Payment Method that is supported in the country_ (otherwise, it won't be displayed).

Please refer to the [Stripe documentation](https://stripe.com/docs/payments/payment-methods/integration-options#additional-api-bank-debits) to better understand all the settings needed to display and use a specific Stripe payment method.

## Checklist

### DIFF
--- a/app/models/solidus_stripe/payment_intent.rb
+++ b/app/models/solidus_stripe/payment_intent.rb
@@ -5,17 +5,9 @@ module SolidusStripe
     belongs_to :order, class_name: 'Spree::Order'
     belongs_to :payment_method, class_name: 'SolidusStripe::PaymentMethod'
 
-    def self.prepare_for_payment(payment, **stripe_intent_options)
+    def self.prepare_for_payment(payment, **stripe_creation_options)
       # Find or create the intent for the payment.
-      intent =
-        retrieve_last_usable_intent(payment) ||
-          new(payment_method: payment.payment_method, order: payment.order)
-            .tap { _1.update!(stripe_intent_id: _1.create_stripe_intent(**stripe_intent_options).id) }
-
-      # Update the intent with the previously acquired payment method.
-      intent.payment_method.gateway.request {
-        Stripe::PaymentIntent.update(intent.stripe_intent_id, payment_method: payment.source.stripe_payment_method_id)
-      }
+      intent = retrieve_for_payment(payment) || create_for_payment(payment, **stripe_creation_options)
 
       # Attach the payment intent to the payment.
       payment.update!(response_code: intent.stripe_intent.id)
@@ -23,9 +15,26 @@ module SolidusStripe
       intent
     end
 
-    def self.retrieve_last_usable_intent(payment)
+    def self.create_for_payment(payment, **stripe_intent_options)
+      new(payment_method: payment.payment_method, order: payment.order)
+        .tap { _1.update!(stripe_intent_id: _1.create_stripe_intent(payment, **stripe_intent_options).id) }
+    end
+
+    def self.retrieve_for_payment(payment)
       intent = where(payment_method: payment.payment_method, order: payment.order).last
-      intent if intent&.usable?
+
+      return unless intent&.usable?
+
+      # Update the intent with the previously acquired payment method.
+      intent.payment_method.gateway.request {
+        Stripe::PaymentIntent.update(
+          intent.stripe_intent_id,
+          payment_method: payment.source.stripe_payment_method_id,
+          payment_method_types: [payment.source.stripe_payment_method.type],
+        )
+      }
+
+      intent
     end
 
     def usable?
@@ -89,7 +98,7 @@ module SolidusStripe
       super
     end
 
-    def create_stripe_intent(**stripe_intent_options)
+    def create_stripe_intent(payment, **stripe_intent_options)
       stripe_customer_id = SolidusStripe::Customer.retrieve_or_create_stripe_customer_id(
         payment_method: payment_method,
         order: order
@@ -102,6 +111,8 @@ module SolidusStripe
           capture_method: payment_method.auto_capture? ? 'automatic' : 'manual',
           setup_future_usage: payment_method.preferred_setup_future_usage.presence,
           customer: stripe_customer_id,
+          payment_method: payment.source.stripe_payment_method_id,
+          payment_method_types: [payment.source.stripe_payment_method.type],
           metadata: { solidus_order_number: order.number },
           **stripe_intent_options,
         })
--- a/app/models/solidus_stripe/payment_intent.rb
+++ b/app/models/solidus_stripe/payment_intent.rb
@@ -59,6 +59,8 @@ module SolidusStripe
       payment.started_processing!
 
       case stripe_intent.status
+      when 'processing'
+        successful = true
       when 'requires_capture'
         payment.pend! unless payment.pending?
         successful = true
--- a/lib/generators/solidus_stripe/install/templates/app/views/checkouts/payment/_stripe.html.erb
+++ b/lib/generators/solidus_stripe/install/templates/app/views/checkouts/payment/_stripe.html.erb
@@ -1,22 +1,57 @@
-<% payment_element_options = {
-  mode: 'payment',
-  amount: SolidusStripe::MoneyToStripeAmountConverter.to_stripe_amount(
-    current_order.display_order_total_after_store_credit.money.fractional,
-    current_order.currency,
-  ),
-  currency: current_order.currency.downcase,
-  paymentMethodCreation: 'manual',
-  captureMethod: payment_method.auto_capture? ? 'automatic' : 'manual',
-  # Fully customizable with appearance API.
-  # https://stripe.com/docs/elements/appearance-api
-  # appearance: {},
-} %>
+<%
+  bill_address = current_order.bill_address
+  options = {
+    elementsInitialization: {
+      mode: 'payment',
+      amount: SolidusStripe::MoneyToStripeAmountConverter.to_stripe_amount(
+        current_order.display_order_total_after_store_credit.money.fractional,
+        current_order.currency,
+      ),
+      currency: current_order.currency.downcase,
+      paymentMethodCreation: 'manual',
+      captureMethod: payment_method.auto_capture? ? 'automatic' : 'manual',
+
+      # If you want to allow the display/use of specific payment method types
+      # activated from the Stripe Dashboard, you can use this option.
+      #
+      # For a complete list of payment method types available on Stripe:
+      # https://stripe.com/docs/payments/payment-methods/overview
+      # https://stripe.com/docs/api/payment_methods/object#payment_method_object-type
+      #
+      # To activate/deactivate Stripe payment methods:
+      # https://dashboard.stripe.com/settings/payment_methods
+      # paymentMethodTypes: ['card', 'sepa_debit'],
+
+      # Fully customizable with appearance API.
+      # https://stripe.com/docs/elements/appearance-api
+      # appearance: {},
+    },
+    paymentElementCreation: {
+      layout: "tabs",
+      defaultValues: {
+        billingDetails: {
+          name: bill_address.name,
+          email: current_order.email,
+          phone: bill_address.phone,
+          address: {
+            line1: bill_address.address1,
+            line2: bill_address.address2,
+            city: bill_address.city,
+            state: bill_address.state.abbr,
+            country: bill_address.country.iso,
+            postal_code: bill_address.zipcode,
+          }
+        }
+      },
+    },
+  }
+ %>
 
 <div
   class="solidus-stripe-payment"
   data-controller="solidus-stripe-payment"
   data-solidus-stripe-payment-publishable-key-value="<%= payment_method.preferred_publishable_key %>"
-  data-solidus-stripe-payment-payment-element-options-value="<%= payment_element_options.to_json %>"
+  data-solidus-stripe-payment-options-value="<%= options.to_json %>"
   data-solidus-stripe-payment-radio-selector-value="#order_payments_attributes__payment_method_id_<%= payment_method.id %>"
   data-solidus-stripe-payment-submit-selector-value="#checkout_form_payment [type='submit']"
   data-action="submit@window->solidus-stripe-payment#handleSubmit"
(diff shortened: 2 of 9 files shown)


app/controllers/solidus_stripe/webhooks_controller.rb
```diff
@@ -5,24 +5,33 @@
...
   class WebhooksController < Spree::BaseController
-    SIGNATURE_HEADER = "HTTP_STRIPE_SIGNATURE"
+    before_action :load_payment_method
 
     skip_before_action :verify_authenticity_token, only: :create
 
     respond_to :json
 
+    rescue_from Webhook::Event::Error do |error|
+      logger.error(error.inspect)
```
kennyadsl: If we move this to a method instead of a block, we leave the possibility to people to change the behavior in case it's needed. Not blocking!

app/controllers/solidus_stripe/webhooks_controller.rb
```diff
@@ -5,24 +5,33 @@
 
 module SolidusStripe
   class WebhooksController < Spree::BaseController
-    SIGNATURE_HEADER = "HTTP_STRIPE_SIGNATURE"
```
kennyadsl: I'm not against this, but is there a specific reason to remove this constant?

kennyadsl (approved):

app/models/solidus_stripe/payment_intent.rb
```diff
@@ -5,27 +5,36 @@ class PaymentIntent < ApplicationRecord
     belongs_to :order, class_name: 'Spree::Order'
     belongs_to :payment_method, class_name: 'SolidusStripe::PaymentMethod'
 
-    def self.prepare_for_payment(payment, **stripe_intent_options)
+    def self.prepare_for_payment(payment, **create_stripe_options)
```
elia: non-blocking nit: `create_stripe_options` → `stripe_creation_options`, just a proposal for your consideration

app/models/solidus_stripe/payment_intent.rb
```diff
@@ -89,7 +98,7 @@ def reload(...)
       super
     end
 
-    def create_stripe_intent(**stripe_intent_options)
+    def create_stripe_intent(payment, **stripe_intent_options)
```
elia, line 103: nit: should this be a private method or maybe inlined?

lib/generators/solidus_stripe/install/templates/app/views/checkouts/payment/_stripe.html.erb
```diff
@@ -1,22 +1,47 @@
...
+    elementsInitialization: {
+      mode: 'payment',
+      amount: SolidusStripe::MoneyToStripeAmountConverter.to_stripe_amount(
+        current_order.display_order_total_after_store_credit.money.fractional,
+        current_order.currency,
+      ),
+      currency: current_order.currency.downcase,
+      paymentMethodCreation: 'manual',
+      captureMethod: payment_method.auto_capture? ? 'automatic' : 'manual',
+      # paymentMethodTypes: ['card', 'sepa_debit'],
```
elia, line 23: nit: Thoughts on adding a short comment above this line indicating it's an example of customizing the method types list?

elia (approved):
Left a few nits, overall looks great! 🙌

app/controllers/solidus_stripe/webhooks_controller.rb
rainerdema: To be honest, no, it's just one less thing in the public API, but happy to revert this change 👍

app/models/solidus_stripe/payment_intent.rb
rainerdema: Moving this method to private involves some changes 🤔 
It's called in the instance created here:
https://github.com/solidusio/solidus_stripe/blob/bb262eb8e282f77cc6d1ca55f6c3e2a37e6500a3/app/models/solidus_stripe/payment_intent.rb#L20
And some options need some other adjustments because they come from instance methods and associations.

I'd rather leave it public, but let me know what you think 👍

app/controllers/solidus_stripe/webhooks_controller.rb
kennyadsl: If that's the concern, we can mark it as private as well. Your call.

app/models/solidus_stripe/payment_intent.rb
elia: That's ok then 👍

waiting-for-dev (changes requested):
Thanks, @rainerdema!

Can we remove the first commit about moving the payment method loading to the controller? As far as I can see:

- It's unrelated and unnecessary to the intention of this PR.
- We're changing to use exceptions to control the flow of execution. That's less performant, and the indirection makes it more difficult to reason about.
- We're moving business logic (how to load a payment method from a slug) to the controller layer.
- We're diluting the separation of concerns. `slug` is a  request thing, so it was ok to be provided to the `.from_request` method.
- More complex. We're now dealing with the unhappy path in two places.
- We already discussed that in https://github.com/solidusio/solidus_stripe/pull/213#discussion_r1131408586, so if we decide to change it I'd like it to have proper visibility so that we can discuss about it properly :pray:

rainerdema:
@waiting-for-dev I will remove the commit as:
- I want to merge this PR as soon as possible 😅 
- This is not a change essential to reach the goal of this PR
- I don't think it's worth spending time right now deciding what might be the best solution
- In addition, as mentioned, the topic has already been covered in another PR 🙏 

Anyway, here are some explanations of why I made the mentioned changes 👀 

>It's unrelated and unnecessary to the intention of this PR.

The main problem about why I started this change was while implementing the sepa_debit payment method, I had some problems figuring out what was wrong with my webhook configuration and the reason for receiving 400 errors from the webhook controller (I was waiting for a "payment_intent.succeeded" from Stripe but I simply had a wrong signature configured).

>We're changing to use exceptions to control the flow of execution. That's less performant, and the indirection makes it more difficult to reason about.

Yep, I can agree about the performance. But not raising anything in case of an exception still makes it difficult to understand what is wrong in case of problems. Maybe some kind of debug mode could be the "solution"?

>We're moving business logic (how to load a payment method from a slug) to the controller layer.
>We're diluting the separation of concerns. slug is a request thing, so it was ok to be provided to the .from_request method.

I don't have much to add here, conceptually, I agree 👍 

>More complex. We're now dealing with the unhappy path in two places.

What do you mean with more complex?

>We already discussed that in https://github.com/solidusio/solidus_stripe/pull/213#discussion_r1131408586, so if we decide to change it I'd like it to have proper visibility so that we can discuss about it properly 🙏

I completely missed this 🙏 

---
### Note

Since the [SolidusStripe::IntentsController.load_payment_method](https://github.com/solidusio/solidus_stripe/blob/100b9c3c22b44dd0625ffdcf045c3845e7462b9e/app/controllers/solidus_stripe/intents_controller.rb#L44) inspired me to load the resource like this, I think we should decide sooner or later: as they are implemented inconsistently with each other currently.

---
### Update

Since there was a force push to remove the commit, you can find the changes mentioned [here](https://github.com/solidusio/solidus_stripe/compare/dceeb026dfb60e680c4f0bc696785065b3ba80e4..55fcde352a9d8e9efbda7ab25b4ce22180323581).

waiting-for-dev (approved):
Thanks, @rainerdema. I think your work doing Q&A on the extension has been super-valuable!!

I'll get back to your comment as soon as possible.

waiting-for-dev:
> The main problem about why I started this change was while implementing the sepa_debit payment method, I had some problems figuring out what was wrong with my webhook configuration and the reason for receiving 400 errors from the webhook controller (I was waiting for a "payment_intent.succeeded" from Stripe but I simply had a wrong signature configured).
> 
> > We're changing to use exceptions to control the flow of execution. That's less performant, and the indirection makes it more difficult to reason about.
> 
> Yep, I can agree about the performance. But not raising anything in case of an exception still makes it difficult to understand what is wrong in case of problems. Maybe some kind of debug mode could be the "solution"?

Yeah, I agree. The reason I think it's not a good idea to get back to Stripe with 5xx error codes is that I'd like to differentiate "expected" failure responses to Stripe (the ones we're handling), vs. unexpected ones (something going wrong in the subscribers). We should have some logging for the expected failures.

> What do you mean with more complex?

With the other solution, both the controller and the `.from_request` method from the `Webhook` class are controlling the unhappy path by raising an exception.

> I completely missed this pray

No worries!


> ### Note
> Since the [SolidusStripe::IntentsController.load_payment_method](https://github.com/solidusio/solidus_stripe/blob/100b9c3c22b44dd0625ffdcf045c3845e7462b9e/app/controllers/solidus_stripe/intents_controller.rb#L44) inspired me to load the resource like this, I think we should decide sooner or later: as they are implemented inconsistently with each other currently.

Yeah, I already shared my concerns in this [comment](https://github.com/solidusio/solidus_stripe/pull/271#discussion_r1162253421). I'd like to see that encapsulated.


I also want to make clear that raising exceptions vs. where we load the payment method are two unrelated things. We can still load the payment method from the `Webhook` class, and raise instead of returning `nil`.

I'm happy to follow the conversation in a dedicated PR!! :slightly_smiling_face:

PR #308, opened 2023-05-23
Allow Solidus v4

elia:
## Summary

Ensure v4 works with Solidus 4 and prep for a new minor release to allow existing installations to upgrade to the latest solidus.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,6 +17,7 @@ jobs:
       ruby_version: "3.2"
     environment:
       FRONTEND: "/home/circleci/project/spec/solidus_frontend_app_template.rb"
+      AUTHENTICATION: "devise"
     steps:
       - browser-tools/install-chrome
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ jobs:
   run-specs-with-postgres:
     executor:
       name: solidusio_extensions/postgres
-      ruby_version: "3.0"
+      ruby_version: "3.2"
     steps:
       - browser-tools/install-chrome
       - checkout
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -26,6 +28,8 @@ jobs:
     executor:
       name: solidusio_extensions/mysql
       ruby_version: "3.1"
+    environment:
+      FRONTEND: "/home/circleci/project/spec/solidus_frontend_app_template.rb"
```
kennyadsl, line 33: Can you please better help me understand why this is needed?

.circleci/config.yml
```diff
@@ -17,6 +17,7 @@ jobs:
       ruby_version: "3.2"
     environment:
       FRONTEND: "/home/circleci/project/spec/solidus_frontend_app_template.rb"
+      AUTHENTICATION: "devise"
```
kennyadsl, line 20: Did we use to test this extension with solidus_auth_devise?

kennyadsl (commented):
Thanks @elia! I left a couple of questions before formal approval.

.circleci/config.yml
elia: Yes, that's because up to Solidus v3.4 solidus_frontend was available as an option, and by just adding the gem to the gemfile we were able to trick the solidus installer into using it, having removed it in v4 we need a custom app template to install it. Rather than having different behaviors for v3 and v4 we just use the same recipe (which is fairly simple anyway).

.circleci/config.yml
elia: Yes, it is necessary for triggering saving a customer on stripe within some specs. I had to force it because solidus main wasn't picking it up anymore (probably related to solidus_frontend support being removed).

.circleci/config.yml
kennyadsl: Got it. Sounds like something we will need to use for all extensions that were testing against the old frontend at some point.

kennyadsl (approved):

rainerdema (approved):
👏

PR #310, opened 2023-05-25
Allow finalizing an order if fully covered by store credits

### LINKED ISSUE #307: Error displaying stripe form when store credits > order.total
When running solidus stripe master branch locally:
- I add 50$ of store credits
- I order for 30$ of products
- so the total to be paid is 0 because everything is taken from the store credits.
- The Stripe checkout form in the frontend checkout display an error 
```
IntegrationError: Invalid value for elements.create(): `amount` must be greater than 0
```
Which seems legit as Stripe probably does not allow a payment for an amout of 0.

Screenshot 2023-05-17 at 10 16 06

**Expected behavior**

it should probably not display the payment forms at all. just "save and continue"

**additional notes**
Everything works fine when store credit amount < order total.
for example with an order total of 94$ and 50$ store credit, the payment workflow looks good
Screenshot 2023-05-17 at 10 29 06

rainerdema:
## Summary

Fixes #307

This temporary fix enables the finalization of orders entirely covered by store credits or adjustments, but it can't be a permanent solution.

The main "issue" relates to Stirpe, which does not allow a 0$ charge because the minimum amount needed to perform the operation isn't met:
https://stripe.com/docs/currencies#minimum-and-maximum-charge-amounts

Various alternative strategies could be followed:
- One option could be to hide the [payment method from the frontend interface](https://github.com/solidusio/solidus_starter_frontend/blob/b9a02e197e8e6ea0aa0b1c9695a18d2b010e21c6/templates/app/views/checkouts/steps/_payment_step.html.erb#L35).
- Map the minimum amounts of all the mentioned currencies and perform a payment with that amount in case of orders fully covered by store credits/adjustments
- Another possibility is to integrate a structure into the `Spree::PaymentMethod` that can prevent a payment method from appearing among the choices available for an order in `available_payment_methods`.

ℹ️ Touching the [`payment_required?`](https://github.com/solidusio/solidus/blob/474877aaad9139a0a9b8564a9e7279e5446bc12c/core/app/models/spree/order.rb#L244-L246) method on `Spree::Order` is something I didn't consider as it would prevent the processing of payments and possibly even the applied store credits, but I haven't looked into that so much, so let me know.

## Checklist

# Screenshots
May-25-2023 17-29-00
Screenshot 2023-05-25 at 17 29 31

### DIFF
--- a/lib/generators/solidus_stripe/install/templates/app/javascript/controllers/solidus_stripe_payment_controller.js
+++ b/lib/generators/solidus_stripe/install/templates/app/javascript/controllers/solidus_stripe_payment_controller.js
@@ -11,6 +11,7 @@ export default class extends Controller {
     // and we can't use outlets, so we fallback on acquiring selectors.
     submitSelector: String,
     radioSelector: String,
+    isCoveredByStoreCredit: Boolean,
   }
 
   static targets = ["paymentElement", "message", "paymentMethodInput"]
@@ -24,8 +25,12 @@ export default class extends Controller {
   }
 
   async connect() {
-    this.stripe = await loadStripe(this.publishableKeyValue)
-    this.setupPaymentElement()
+    if (this.isCoveredByStoreCreditValue) {
+      this.messageTarget.textContent = "Free orders are not supported."
+    } else {
+      this.stripe = await loadStripe(this.publishableKeyValue)
+      this.setupPaymentElement()
+    }
   }
 
   // @action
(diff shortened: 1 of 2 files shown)


rainerdema:
@solidusio/core-team thoughts on https://github.com/solidusio/solidus_stripe/pull/310#issue-1725969255?

waiting-for-dev:
I think, ideally, our business logic should be able to move forward when there's no required money. If that's not possible, then we should tackle it at the view layer, but I wouldn't enforce a minimum amount.

rainerdema:
This will be addressed with: [Preventing available payment method rendering for orders fully covered by Store Credits #345
](https://github.com/solidusio/solidus_starter_frontend/issues/345)

PR #318, opened 2023-11-29
Bug Fix: Register Webhook Event Only Once

cpfergus1:
Prior to this change, an error would be thrown every time Rails would have to reload because SolidusStripe would try to re-register the webhook events. To prevent this we want to remove the registration call outside of the reloader so it only registers the event once.

### DIFF
--- a/lib/solidus_stripe/engine.rb
+++ b/lib/solidus_stripe/engine.rb
@@ -18,11 +18,13 @@ module SolidusStripe
 
     initializer "solidus_stripe.pub_sub", after: "spree.core.pub_sub" do |app|
       require "solidus_stripe/webhook/event"
+
+      SolidusStripe::Webhook::Event.register(
+        user_events: SolidusStripe.configuration.webhook_events,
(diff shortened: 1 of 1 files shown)


lib/solidus_stripe/engine.rb
```diff
@@ -18,11 +18,13 @@ class Engine < Rails::Engine
...
+        user_events: SolidusStripe.configuration.webhook_events,
+        bus: Spree::Bus
+      )
+
       app.reloader.to_prepare do
-        SolidusStripe::Webhook::Event.register(
-          user_events: SolidusStripe.configuration.webhook_events,
-          bus: Spree::Bus
-        )
         SolidusStripe::Webhook::PaymentIntentSubscriber.new.subscribe_to(Spree::Bus)
```
kennyadsl, line 28: Did you investigate if we can also move the subscribers out of the `to_prepare` block? I'm not sure but maybe we are subscribing to events multiple times if we keep them in?

elia (approved):

PR #319, opened 2023-11-30
Revert "Bug Fix: Register Webhook Event Only Once"

elia:
Reverts solidusio/solidus_stripe#318

Reverting since the CI failed on `main`, we need to look into this further.

cc @cpfergus1

### DIFF
--- a/lib/solidus_stripe/engine.rb
+++ b/lib/solidus_stripe/engine.rb
@@ -18,13 +18,11 @@ module SolidusStripe
 
     initializer "solidus_stripe.pub_sub", after: "spree.core.pub_sub" do |app|
       require "solidus_stripe/webhook/event"
-
-      SolidusStripe::Webhook::Event.register(
-        user_events: SolidusStripe.configuration.webhook_events,
-        bus: Spree::Bus
-      )
-
       app.reloader.to_prepare do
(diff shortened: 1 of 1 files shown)


cpfergus1:
@elia 
I found two other ways to fix this, but I am still failing some tests locally and would be interested to see what CI returns (wondering if it is my environment), LMKWYT!

```
lib/solidus_stripe/webhook/event.rb
       def register(user_events:, bus:, core_events: CORE_EVENTS)
          (core_events + user_events).each do |event|
            if bus.registry.registrations.select { |registration| registration.event_name == :"stripe.#{event}" }.empty? 
              bus.register(:"stripe.#{event}")
            end
          end
        end
```

```
lib/solidus_stripe/engine.rb
    initializer "solidus_stripe.pub_sub", after: "spree.core.pub_sub" do |app|
      require "solidus_stripe/webhook/event"

      app.reloader.to_prepare do
        SolidusStripe::Webhook::Event.register(
          user_events: SolidusStripe.configuration.webhook_events,
          bus: Spree::Bus
        )
        SolidusStripe::Webhook::PaymentIntentSubscriber.new.subscribe_to(Spree::Bus)
        SolidusStripe::Webhook::ChargeSubscriber.new.subscribe_to(Spree::Bus)
      end
    end
```

PR #322, opened 2024-02-02
Add partial support for v4 intents payment sources

rabbitbike:
## Summary

This PR tries to add support for the continued use of v4 payment sources saved in `wallet_payment_sources` as "existing payments." The code is more of a proof of concept so I want to present this PR as a draft.

Currently v4 payment sources cannot be reused and this makes it difficult for stores to migrate to v5 if they have already implemented v4 in production. This PR aims to resolve that. 

I believe the support is only applicable for stores that used Stripe Payment Intents API with the `v3_intents` option enabled, but I'm not entirely sure.
It does not support creation of new v4 payment sources.
It only allows reuse of cards saved in `wallet_payment_sources`.

I am well aware that v5 is a complete rewrite, and I believe the core team doesn't want code supporting legacy versions creeping into various parts of the code base. Still, I wanted to see if there is any way something like this can be implemented at the gem level. Also, I would like to kindly ask if the reuse of v4 payment sources are done in a correct way. 

I hope this PR can provide easier migration path for existing users of this wonderful gem.

Thank you for reviewing this PR and I appreciate any comments or suggestions🙏

## Update

I realized that what I described in this PR requires the `payment_method_id` to be the same for both v4 and v5. So this PR will not work when a payment method is added instead of updated from v4.

## Checklist

- [X] I have written a thorough PR description.
- [X] I have kept my commits small and atomic.
- [X] [I have used clear, explanatory commit messages](https://github.com/solidusio/.github/blob/main/CONTRIBUTING.md#writing-good-commit-messages).

### DIFF
--- a/app/decorators/models/solidus_stripe/spree/credit_card_decorator.rb
+++ b/app/decorators/models/solidus_stripe/spree/credit_card_decorator.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+module SolidusStripe
+  module Spree
+    module CreditCard
+      # Support for partial backward compatibility with solidus_stripe v5 if Payment Intents API was used in v4
+      # Allows continued use of v4 credit cards saved in user's wallet_payment_sources
+      # Cannot create a new v4 payment source
+
+      def stripe_customer_id
+        gateway_customer_profile_id
+      end
+
+      def stripe_payment_method_id
+        gateway_payment_profile_id
+      end
+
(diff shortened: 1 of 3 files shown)


rabbitbike:
I would like to close this PR as this will not work in many situations, including the one I described when you upgrade to v5 with a new payment_method with a new id.

I hope when there is a good migration strategy it will be documented.

Thank you🙏

PR #327, opened 2024-07-29
Implement API

kchodorek:
## Summary

Recently, we've been working on a Solidus-based project that uses Solidus via an API. To enable Stripe payments within this setup, we needed to add additional API endpoints and logic for creating payment intents within Stripe. The intent information can later be used in combination with the [Stripe Payment Element](https://docs.stripe.com/payments/payment-element).

This PR aims to integrate these changes into the upstream `solidus_stripe`.

It adds the following API endpoints:

- `Create Setup Intent`
- `Create Payment Intent`

More details on these endpoints can be found in the README [here](https://github.com/kchodorek/solidus_stripe/blob/a134d072ada24b3e606b7db309874a64241bc6c1/README.md#api-support).

The core logic of this change is encapsulated in service objects, making it easy to create these intents outside of the API as well (e.g., if you prefer a custom GraphQL mutation or need to create a payment intent outside of the regular checkout process for some reason).

## Checklist

### DIFF
--- a/app/models/solidus_stripe/customer.rb
+++ b/app/models/solidus_stripe/customer.rb
@@ -19,9 +19,7 @@ module SolidusStripe
     end
 
     def customer_name
-      return full_name if full_name.present?
-
-      source.email
+      full_name.presence || source.email
     end
 
     private
--- a/app/models/solidus_stripe/customer.rb
+++ b/app/models/solidus_stripe/customer.rb
@@ -15,7 +15,21 @@ module SolidusStripe
     end
 
     def create_stripe_customer
-      payment_method.gateway.request { Stripe::Customer.create(email: source.email) }
+      payment_method.gateway.request { Stripe::Customer.create(email: source.email, name: customer_name) }
+    end
+
+    def customer_name
+      return full_name if full_name.present?
+
+      source.email
+    end
+
+    private
+
+    def full_name
+      return unless source.respond_to?(:first_name) && source.respond_to?(:last_name)
+
+      "#{source.first_name} #{source.last_name}".strip
     end
   end
 end
--- a/bin/sandbox
+++ b/bin/sandbox
@@ -15,6 +15,7 @@ function unbundled {
 
 cd "./$app_name"
 unbundled bundle add $extension_name --path ..
+unbundled bundle add stripe-ruby-mock --group "test"
 unbundled bundle exec rails generate $extension_name:install --migrate --specs=all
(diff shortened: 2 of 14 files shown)


app/models/solidus_stripe/customer.rb
```diff
@@ -15,7 +15,21 @@ def self.retrieve_or_create_stripe_customer_id(payment_method:, order:)
...
 
     def create_stripe_customer
-      payment_method.gateway.request { Stripe::Customer.create(email: source.email) }
+      payment_method.gateway.request { Stripe::Customer.create(email: source.email, name: customer_name) }
+    end
+
+    def customer_name
+      return full_name if full_name.present?
+
+      source.email
```
jarednorman: This can be written simply as `full_name.presence || source.email` if you prefer.

jarednorman (commented):
There's _a lot_ of mocking in these tests. I only skimmed this, but I didn't see anything resembling an integration test that tests if these new endpoints actually work as intended. 

I'll need to take a more serious look, but I'm concerned about the maintainability of this as is. I'd much prefer something where the controller tests aren't stubbing everything out and instead test that the objects work as arranged. I'd also try to pull the object construction up higher so there's less of the constructor stubbing code smell (`allow(SomeClass).to receive(:new)...`).

kchodorek:
@jarednorman Thanks for your comment! I adjusted the specs, now instead of stubbing, I'm using `stripe-ruby-mock` gem, tests are much cleaner now.

bin/sandbox
```diff
@@ -15,6 +15,7 @@ function unbundled {
 
 cd "./$app_name"
 unbundled bundle add $extension_name --path ..
+unbundled bundle add stripe-ruby-mock --group "test"
```
kennyadsl: Can you help me remember why we need this in the sandbox? I though having the dependency. in the dummy-app would be enough, but maybe I'm missing something.

kennyadsl (commented):
Thanks for the contribution, left a comment but overall looks good! ❤️

bin/sandbox
kchodorek: I wasn't sure if sandbox also suppose to include the new dependency - in such case I will remove it from `sandbox` and will leave it in `dummy-app`, thanks!

jarednorman (approved):
Thanks, that's much better.

kennyadsl (approved):

PR #331, opened 2026-06-29
Fix rails-new binstub

adammathys:
## Summary

Brings it more inline with what we're using in other extensions. Important updates to include the manifest.js file and postgres config so that our tests can actually run again.


adammathys:
Closed in favour of https://github.com/solidusio/solidus_stripe/pull/334

PR #332, opened 2026-06-29
Bump to Stripe v13

adammathys:
## Summary

Bumps the version of the Stripe gem to '~> 13.0' so that we have access to more modern features of the Stripe API. This is also the newest version we can use that still supports the now deprecated `StripeClient#request` method that we're using for basically all of our API calls. That method is deprecated in favour a service based call pattern or using `StripeClient#raw_request` which has a different interface.

The only version with any migration requirements before v13 is v9, which introduced a change to always sending the API version header with requests. None of the parameters or APIs we're using appear to be impacted by that change so it's okay for us to just let it start sending the API version for us. If customers were locked to an older version of the API, they can skill explicitly set the API by setting it in an initializer like so: `Stripe.api_version = '2020-08-27'`

With the extension on this version, we can slowly migrate our code to the new service based call pattern and eventually upgrade to the latest version of the Stripe gem.

https://github.com/stripe/stripe-ruby/wiki/Migration-guide-for-v13


AlistairNorman:
Closing since this work has been combined with #336

